PR #20, opened 2020-02-27
19 port 8088 is blocked

jrob:
It appears that 8088 is blocked unless you have enterprise set.


naseemkullah:
I don't quite get why, but it's always worked (backup cron jobs which talk to the influxdb on port 8088 succeed). 🤷‍♂

naseemkullah:
cc @rawkode @gitirabassi @Aisuko

rawkode:
Yep. You are correct. This is because the service uses a named port. Good catch

PR #31, opened 2020-03-02
[telegraf-ds] Improve documentation

nsteinmetz:
* Improve documentation 
* fix some helm commands to be compatible with helm 3


charts/telegraf-ds/README.md
```diff
@@ -46,7 +46,7 @@ The command removes all the Kubernetes components associated with the chart and
 The default configuration parameters are listed in `values.yaml`. To change the defaults, specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example,
 
 ```console
-helm install --name my-release \
+helm install my-release \
```
rawkode: Maybe it's better to use `helm upgrade --install` instead of a one time `helm install` command; as the `helm upgrade --install` is idempotent and can be run as many times as needed.

nsteinmetz: I agree and it would also manage the "real" upgrade case. I'll apply later - do you want me to apply it also on all other charts ?

rawkode: We can add an issue for the other charts and keep this PR on DS :+1:

nsteinmetz: Ok, just pushed the fix for the DS only

PR #32, opened 2020-03-03
Use of helm upgrade --install and helm uninstall

nsteinmetz:
* all charts to use `helm upgrade --install` instead of `helm install`
* `helm delete --purge` replaced by `helm uninstall`


rawkode:
I don't think the chart versions need bumped for the README, but I don't mind the bump either way

nsteinmetz:
Ok, I thought each update would require a bump - will know it for next time ;-)

PR #33, opened 2020-03-03
Update values.yaml

IvanKirianov:
It looks like an issue, based on _helpers.tpl, should be only one "-". 
This causes an error and influxdb setup with the localhost instead of filled value.


nsteinmetz:
Oh yes indeed, my bad - a bad and quick copy/paste - sorry for that.

You need to change the `version` in `Chart.yaml` to be > 1.0.1

nsteinmetz:
LGTM - @rawkode could you please review it ?

PR #34, opened 2020-03-04
fix: bump version of telegraf-ds due to merging 2 PRs with same version

rawkode:


nsteinmetz:
ah ah so even for an update of the readme, the bump version is required - we have our answer now ;-)

PR #35, opened 2020-03-05
feat: Add InfluxDB2 Chart

rawkode:


charts/influxdb2/templates/job-setup-admin.yaml
```diff
@@ -0,0 +1,48 @@
...
+        args:
+          - -c
+          - |
+            influx setup -f \
+            --host http://{{ template "influxdb.fullname" . }}:{{ .Values.service.port }} \
+            -o {{ .Values.adminUser.organization }} \
+            -b {{ .Values.adminUser.bucket }} \
+            -u {{ .Values.adminUser.user }} \
+            -p ${INFLUXDB_PASSWORD} \
+            -t $INFLUXDB_TOKEN
```
goller: I think that `-t $INFLUXDB_TOKEN` may not be needed; I glanced quickly at the code and I believe -t reads that environment variable.

charts/influxdb2/templates/statefulset.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: apps/v1beta2
+kind: Deployment
```
goller: File name is statefulset, but, the kind is Deployment.

I think it makes sense for it to be a deployment rather than a statefulset.

charts/influxdb2/templates/job-setup-admin.yaml
russorat: in think the env variable is actually INFLUX_TOKEN: https://github.com/influxdata/influxdb/issues/16048#issuecomment-562664102

charts/influxdb2/templates/statefulset.yaml
gitirabassi: I would argue that it make sense to be a statefulset

charts/influxdb2/values.yaml
```diff
@@ -0,0 +1,55 @@
...
+  ## influxdb data Persistent Volume Storage Class
+  ## If defined, storageClassName: <storageClass>
+  ## If set to "-", storageClassName: "", which disables dynamic provisioning
+  ## If undefined (the default) or set to null, no storageClassName spec is
+  ##   set, choosing the default provisioner.  (gp2 on AWS, standard on
+  ##   GKE, AWS & OpenStack)
+  ##
+  # storageClass: "-"
+  accessMode: ReadWriteOnce
+  size: 8Gi
```
gitirabassi: Let's provide a default of `50Gi` at least

charts/influxdb2/values.yaml
```diff
@@ -0,0 +1,55 @@
+image:
+  repository: quay.io/influxdb/influxdb
+  pullPolicy: IfNotPresent
+
+nameOverride: ""
+fullnameOverride: ""
+
+service:
+  type: ClusterIP
+  port: 80
```
gitirabassi: shouldn't this be 9999?

naseemkullah:
If you have the bandwidth, running `helm create influxdb-2` (with latest version of helm) will result in better boilerplate (selector label helpers, serviceaccount, image tag equal to chart's app version) and build from there.

rawkode:
I did use `helm create` ☹️

I'll recreate it on Monday and fix up the other suggestions

charts/influxdb2/templates/job-setup-admin.yaml
rawkode: I think I prefer this being explicit and people knowing that we're setting it to the value of an environment variable, rather than it implicitly using it

charts/influxdb2/values.yaml
rawkode: The service will be available on port 80, it redirects to the pods 9999 port

rawkode:
The linting action doesn't work for v2 charts, but I've tested it locally. We'll force merge it once this has been reviewed

PR #36, opened 2020-03-06
[influxdb] Add existingClaim option.

naseemkullah:
The option to provide a pre existing PVC to mount was removed during the move from Deployment to StatefulSet via https://github.com/helm/charts/pull/17619/files. Seeing that influxdb OSS runs as a single replica, mounting a pre existing PVC remains a valid option, especially helpful for data continuity across cluster rebuilds where the user recreates a PVC claiming the original (and retained) PV.

### DIFF
--- a/charts/influxdb/Chart.yaml
+++ b/charts/influxdb/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: influxdb
-version: 4.3.6
+version: 4.4.0
 appVersion: 1.7.10
 description: Scalable datastore for metrics, events, and real-time analytics.
 keywords:
--- a/charts/influxdb/README.md
+++ b/charts/influxdb/README.md
@@ -60,7 +60,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | startupProbe | Health check for pod | {} |
 | service.type | Kubernetes service type | ClusterIP |
 | persistence.enabled | Boolean to enable and disable persistance | true |
-| persistence.existingClaim | An existing PersistentVolumeClaim | nil |
+| persistence.existingClaim | An existing PersistentVolumeClaim, ignored if enterprise.enabled=true | nil |
 | persistence.storageClass | If set to "-", storageClassName: "", which disables dynamic provisioning. If undefined (the default) or set to null, no storageClassName spec is set, choosing the default provisioner.  (gp2 on AWS, standard on GKE, AWS & OpenStack |  |
 | persistence.annotations | Annotations for volumeClaimTemplates | nil |
 | persistence.accessMode | Access mode for the volume | ReadWriteOnce |
--- a/charts/influxdb/README.md
+++ b/charts/influxdb/README.md
@@ -60,6 +60,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | startupProbe | Health check for pod | {} |
 | service.type | Kubernetes service type | ClusterIP |
 | persistence.enabled | Boolean to enable and disable persistance | true |
+| persistence.existingClaim | An existing PersistentVolumeClaim | nil |
 | persistence.storageClass | If set to "-", storageClassName: "", which disables dynamic provisioning. If undefined (the default) or set to null, no storageClassName spec is set, choosing the default provisioner.  (gp2 on AWS, standard on GKE, AWS & OpenStack |  |
 | persistence.annotations | Annotations for volumeClaimTemplates | nil |
 | persistence.accessMode | Access mode for the volume | ReadWriteOnce |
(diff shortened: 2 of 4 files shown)


naseemkullah:
cc @Aisuko

rawkode:
I'm not very keen on this being added

It feels like people could very easily screw up their PV by scaling the StatefulSet

There's also the naming conflict between the existingClaim and the provisionedClaim

I feel that anyone using the StatefulSet should use the provisionedClaims and migrate their data from an existingClaim

naseemkullah:
> I'm not very keen on this being added
> 
> It feels like people could very easily screw up their PV by scaling the StatefulSet
> 

Can one scale up influxdb OSS beyond 1 pod? We can have it disabled if enterprise is enabled.

> There's also the naming conflict between the existingClaim and the provisionedClaim
> 

They can't both be rendered as per the surrounding if statements. If existingClaim is enabled, only it is rendered, if it is disabled, the provisionedClaim is rendered.

rawkode:
The StatefulSet could be scaled up, but it wouldn't be a "cluster"

naseemkullah:
> The StatefulSet could be scaled up, but it wouldn't be a "cluster"

Is this ever a desired approach?

rawkode:
No, but it doesn't mean someone won't do it?

naseemkullah:
So to recap with regards to OSS:

1. exeistingClaim would mess things up for statefulsts larger than 1 replics
2. users should not scale up beyond 1 replica

I don't see an issue.

naseemkullah:
bump

rawkode:
@naseemkullah I won't accept this PR as it currently is.

I've looked through the `helm/charts` repository and it seems that the "Helm way" to provide this option is to make the `StatefulSet`  a `Deployment` when `existingClaim` is configured.

So if `existingClaim` is provided, the `StatefulSet` isn't rendered, and a `Deployment` is.

If you match this, then we can get this merged.

https://github.com/helm/charts/blob/1ccfc8be4f3ca5f26b991f7e1d2eaccd9bbefadf/stable/grafana/templates/statefulset.yaml

naseemkullah:
Fair enough!
Did you want Deployment rendered instead of StatefulSet if persistence is disabled?

rawkode:
@naseemkullah Yes, please. This reduces the risk of us killing some teams PVCs from the StatefulSet

Humans make mistakes, lets protect them as much as possible.

naseemkullah:
> Humans make mistakes, lets protect them as much as possible.

💯

rawkode:
Hmm. I wonder if we should extract the enterprise stuff into `influxdb-enterprise` and simplify the `influxdb` chart to only be OSS

Then this wouldn't be a problem at all ...

naseemkullah:
> Hmm. I wonder if we should extract the enterprise stuff into `influxdb-enterprise` and simplify the `influxdb` chart to only be OSS
> 
> Then this wouldn't be a problem at all ...

I don't mind that approach, but would there be duplicate code? E.g. backup cronjob template.

naseemkullah:
Also on a last note:

```
      # Cannot use existing claim in enterprise mode
      {{- if and .Values.persistence.enabled .Values.persistence.existingClaim (not .Values.enterprise.enabled) }}
      - name: {{ include "influxdb.fullname" . }}-data
        persistentVolumeClaim:
          claimName: {{ .Values.persistence.existingClaim }}
      {{- end }}
  # Must use volume claim template in enterprise mode
  {{- if and .Values.persistence.enabled (or (not .Values.persistence.existingClaim) .Values.enterprise.enabled) }}
  volumeClaimTemplates:
    - metadata:
        name: {{ include "influxdb.fullname" . }}-data
```

FWIW IF we make above change, enterprise mode will NOT be able to use an existingClaim.

rawkode:
Enterprise wouldn't use an existingClaim. It should be using provisioned PVCs from StatefulSets

naseemkullah:
Exactly, and the above would make it impossible to.

naseemkullah:
The if statements have been adjusted as such.

Both `helm template charts/influxdb`(default values) and `helm template charts/influxdb --set persistence.existingClaim=ok,enterprise.enabled=true` (both enterprise and existingClaim enabled) yield:

```yaml
      volumes:
      - name: config
        configMap:
          name: RELEASE-NAME-influxdb
      # Cannot use existing claim in enterprise mode
  # Must use volume claim template in enterprise mode
  volumeClaimTemplates:
    - metadata:
        name: RELEASE-NAME-influxdb-data
        annotations:
      spec:
        accessModes:
          - "ReadWriteOnce"
        resources:
          requests:
            storage: "8Gi"
```

`helm template charts/influxdb --set persistence.existingClaim=ok` (existingClaim enabled, but not enterprise)
```yaml
      volumes:
...
      # Cannot use existing claim in enterprise mode
      - name: RELEASE-NAME-influxdb-data
        persistentVolumeClaim:
          claimName: ok
  # Must use volume claim template in enterprise mode

( no volume Claim templates)
```

naseemkullah:
@gitirabassi Could you please re-approve? Made some changes since to ensure enteprise and existingClaim are never used in conjunction.

PR #38, opened 2020-03-06
[influxdb] Fix config rendering.

naseemkullah:
Provided that int values can be quoted in the config, this fixes the issue where strings are not quoted in the rendered config.

e.g. the value:

```
config:
  http:
    max_row_limit: 0
  logging:
    format: json
```

renders in the configmap, prior to this commit:
```
    [http]
      max_row_limit = 0
[...]
    [logging]
      format = json
```

Which results in the following error: `run: parse config: Near line 43 (last key parsed 'logging.format'): expected value but found "json" instead`

While this commit renders:

```
    [http]
      max_row_limit = "0"
[...]
    [logging]
      format = "json"
```

which produces no error.

### DIFF
--- a/charts/influxdb/Chart.yaml
+++ b/charts/influxdb/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: influxdb
-version: 4.3.4
+version: 4.3.5
 appVersion: 1.7.10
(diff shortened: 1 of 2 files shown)


naseemkullah:
cc @Aisuko

PR #41, opened 2020-03-09
fix(influxdb): pass boolean conf values as is to configmap

hugocortes:
My influxdb `config` object contains some boolean values which was causing this error to be thrown:
```sh
run: parse config: toml: cannot load TOML value of type string into a Go boolean
```

The influxdb configuration requires boolean values to not be quoted and strings to be quoted. With the type check update, a resulting config snippet as follows:
```toml
     [http]
       auth-enabled = true
       bind-address = ":8086"
       bind-socket = "/var/run/influxdb.sock"
       enabled = true
       flux-enabled = true
       https-certificate = "/etc/ssl/influxdb.pem"
       https-enabled = false
       log-enabled = true
```

Snippet of my values:
```yaml
config:
  logging:
    level: warn
  http:
    bind-address: :8086
    auth-enabled: true
    flux-enabled: true
    log-enabled: true
    write-tracing: false
    pprof-enabled: true
    https-enabled: false
    https-certificate: /etc/ssl/influxdb.pem
```

### DIFF
--- a/charts/influxdb/Chart.yaml
+++ b/charts/influxdb/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: influxdb
-version: 4.3.5
+version: 4.3.6
 appVersion: 1.7.10
 description: Scalable datastore for metrics, events, and real-time analytics.
 keywords:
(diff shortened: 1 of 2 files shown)


naseemkullah:
Thanks @hugocortes !

naseemkullah:
cc @gitirabassi @rawkode

PR #44, opened 2020-03-16
add scheme to influxDB healthcheck to serve HTTPS

songrgg:
When influxdb is serving HTTPS on API endpoint, the healthcheck probe will fail because it defaults to HTTP get instead of HTTPS, this merge request is to add an option in values.yml to setup HTTPS healthcheck if influxDB is serving HTTP requests.

### DIFF
--- a/charts/influxdb/Chart.yaml
+++ b/charts/influxdb/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: influxdb
-version: 4.4.1
+version: 4.4.2
 appVersion: 1.7.10
 description: Scalable datastore for metrics, events, and real-time analytics.
 keywords:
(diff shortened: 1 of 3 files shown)


rawkode:
@songrgg Can you bump the version of the chart, please?

songrgg:
@rawkode I'm testing on influxDB 1.7.10

rawkode:
@songrgg Sorry. You need to bump the version in the Chart.yaml otherwise the build won't pass.

Can you increment the 3rd number by 1, please?

songrgg:
Got it, will do that.

songrgg:
@rawkode bump it from 1.7.10 to 1.7.11 :)

rawkode:
Sorry @songrgg ,

I believe I was unclear. Can you bump `version: 4.4.1` to `version: 4.4.2` and revert the `appVersion`?

The reason we bump the `version: x.x.x` is so that the version numbers don't collide when we publish the artefacts.

rawkode:
Thank you @songrgg

PR #47, opened 2020-03-20
[influxdb] Add podAnnotations to backup pods.

naseemkullah:


charts/influxdb/values.yaml
```diff
@@ -228,6 +228,7 @@ backup:
   enabled: false
   schedule: "0 0 * * *"
   annotations: {}
+  podAnnotation: {}
```
rawkode: Think this should be `podAnnotations`

naseemkullah:
> What's the usecase for this?

Use case is to disable sidecar injection in the case of namespace-wide sidecar injection.

charts/influxdb/values.yaml
naseemkullah: yes, good eye!

charts/influxdb/README.md
```diff
@@ -95,9 +95,12 @@ The command removes all the Kubernetes components associated with the chart and
 | config.tls | [Details](https://docs.influxdata.com/influxdb/v1.7/administration/config/#tls) | {} |
 | initScripts.enabled | Boolean flag to enable and disable initscripts. If the container finds any files with the extensions .sh or .iql inside of the /docker-entrypoint-initdb.d folder, it will execute them. The order they are executed in is determined by the shell. This is usually alphabetical order. | false |
 | initScripts.scripts | Init scripts | {} |
-| backup.enabled | Boolean flag to enable and disable backups. Currently, it backups the data on `azure` and `gcs`. | false |
-| backup.schedule | Cron time | `0 0 * * *`. It means create a backup everyday at `00:00`. |
-| backup.annotations | Annotations for backup | {} |
+| backup.enabled | Enable backups, if `true` must configure one of the storage providers | `false` |
+| backup.gcs | Google Cloud Storage config | `nil`
```
gitirabassi: What if neither `backup.gcs` nor `backup.azure` are set?

naseemkullah: We should make a PR to add logic to the backup cronjob to only be created it a storage provider is set! Good call.

PR #55, opened 2020-04-01
Add telegraf-operator chart

gitirabassi:
Closes https://github.com/influxdata/telegraf-operator/issues/7

This will need https://github.com/influxdata/telegraf-operator/milestone/2 to be completed

I need helm configuring this to be proper helm etiquette

What is the proper way of adding a secret with helm? @naseemkullah 

I'm thinking of adding `influxdb` as a dependency and then adding the secret to send telegraf data to influxdb instance. 

This can't be done with influxdbV2 because of all the initialization and token distribution

### DIFF
--- a/charts/telegraf-operator/README.md
+++ b/charts/telegraf-operator/README.md
@@ -1,0 +1,25 @@
+# Telegraf-operator
+
+> Default installation expects cert-manager to be running in the cluster
+
+## Usage
+
+```console
+helm repo add influxdata https://helm.influxdata.com
+
+helm install telegraf-operator influxdata/telegraf-operator
+
+```
+
+## Contributing & Developing
+
+```shell
+helm template --namespace=telegraf-operator telegraf-operator .
+```
+
+Testing CI template
+```shell
+helm template --namespace=telegraf-operator --values=./ci/values.yaml telegraf-operator .
+```
+
+
--- a/charts/telegraf-operator/templates/certificate.yml
+++ b/charts/telegraf-operator/templates/certificate.yml
@@ -1,4 +1,4 @@
-{{- if .Values.certManager.certificate.enable -}}
+{{- if .Values.certManager.enable -}}
 apiVersion: cert-manager.io/v1alpha2
 kind: Certificate
 metadata:
--- a/charts/telegraf-operator/templates/certificate.yml
+++ b/charts/telegraf-operator/templates/certificate.yml
@@ -1,0 +1,18 @@
+{{- if .Values.certManager.certificate.enable -}}
+apiVersion: cert-manager.io/v1alpha2
+kind: Certificate
+metadata:
+  name: {{ include "telegraf-operator.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "telegraf-operator.labels" . | nindent 4 }}
+spec:
+  dnsNames:
+  - "{{ include "telegraf-operator.fullname" . }}.{{ .Values.namespace }}.svc"
+  isCA: true
+  issuerRef:
+    kind: Issuer
+    name: telegraf-operator-issuer
+    namespace: {{ .Release.Namespace }}
+  secretName: telegraf-operator-tls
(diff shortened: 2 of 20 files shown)


rawkode:
Dependencies in Helm are pretty shit. The secret should be provided upfront via a value `secretName` `

nsteinmetz:
Please don't:

> I'm thinking of adding influxdb as a dependency and then adding the secret to send telegraf data to influxdb instance.

InfluxDB instance can be elsewhere or one may want to user another output system than InfluxDB

In my case, my monitoring system is hosted externally from the kubernetes cluster ; I don't want to deploy a influxdb instance which I'll not use.

gitirabassi:
@nsteinmetz Thank you for your feedback. I'm all for customizability but also making sure that people that are getting started find everything they need here.

Just like it's possible to disable the `cert-manager`  dependency now, I'll make `influxdb` dependency optional. 

The question now is: 
- should it be enabled by default
- should it be disabled by default
cc @rawkode

rawkode:
@gitirabassi Disabled by default. I don't like coupling, and we'll want to support the InfluxDB Operator later - so lets not tie this to InfluxDB too much now.

gitirabassi:
@rawkode can you help me understand what is going on with CI ? I can't figure out why it's failing

charts/telegraf-operator/README.md
```diff
@@ -0,0 +1,26 @@
...
+helm repo add influxdata https://helm.influxdata.com
+
+helm install telegraf-operator influxdata/telegraf-operator
+
+```
+
+## Contributing
+
+```
+helm dependency update
```
rawkode, line 17: Is this required?

charts/telegraf-operator/templates/certificate.yml
```diff
@@ -0,0 +1,18 @@
...
+  namespace: {{ .Values.namespace }}
+spec:
+  dnsNames:
+  - "{{ include "telegraf-operator.fullname" . }}.{{ .Values.namespace }}.svc"
+  isCA: true
+  issuerRef:
+    kind: Issuer
+    name: selfsigned
+    namespace: {{ .Values.namespace }}
+  secretName: webhook-server-cert
```
rawkode, line 17: Should this be something like `telegraf-operator-tls` ?

charts/telegraf-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "telegraf-operator.fullname" . }}
+  namespace: {{ .Values.namespace }}
+  labels:
+    {{- include "telegraf-operator.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
```
rawkode, line 9: Can the operator be horizontally scaled, or should we hard code this to 0?

charts/telegraf-operator/templates/issuer.yml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.certManager.certificate.enable -}}
+apiVersion: cert-manager.io/v1alpha2
+kind: Issuer
+metadata:
+  name: selfsigned
```
rawkode, line 5: Should this be a `telegraf-operator-issuer`?

gitirabassi: done

charts/telegraf-operator/templates/deployment.yaml
gitirabassi: it already defaults to 3

charts/telegraf-operator/templates/certificate.yml
gitirabassi: done

charts/telegraf-operator/README.md
gitirabassi: done

gitirabassi:
@rawkode This is now ready for final review

PR #56, opened 2020-04-03
Upgrade telegraf to 1.14 for telegraf & telegraf-ds charts

nsteinmetz:
Upgrade telegraf version to 1.14 for both charts


nsteinmetz:
Sorry @rawkode, just noticed a typo in the telegraf's readme - so I take the fix with this PR instead to open a new one ;-)

PR #62, opened 2020-04-15
Kapacitor sideloads

ydkn:
This PR will allow to load sideload files from config maps. The sideload files will be be placed in /var/lib/kapacitor/sideload by default but is customizable through `sidecar.sideload.folder`.
Please see the updated Readme for all new available values.

This change is based on the functionality to load dashboards in the grafana chart (https://github.com/helm/charts/tree/master/stable/grafana)

Example config map:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: kapacitor-sideload-influxdb-httpd-clienterror
  labels:
    kapacitor_sideload: "1"
data:
  influxdb-httpd-clienterror.yml: |
    [...]
```

Please let me know what you think about this. Since I'm pretty new to Kapacitor I'm open to shape this to additional needs that I'm currently not aware of.

### DIFF
--- a/charts/kapacitor/Chart.yaml
+++ b/charts/kapacitor/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kapacitor
-version: 1.2.10
+version: 1.3.0
 appVersion: 1.5.4
 description: InfluxDB's native data processing engine. It can process both stream
   and batch data from InfluxDB.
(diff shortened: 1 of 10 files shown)


rawkode:
Hi, thank you for your PR. Apologies, but I'm not  sure what use-case this is trying to accomplish.

Can you provide a little more context, please?

ydkn:
If you have a TICKscript and want to overwrite some parameters for a specific host you can integrate a sideload e.g. (taken from https://docs.influxdata.com/kapacitor/v1.5/nodes/sideload_node/)

```
|sideload()
  .source('file:///path/to/dir')
  .order('host/{{.host}}.yml', 'hostgroup/{{.hostgroup}}.yml')
  .field('cpu_threshold', 0.0)
  .tag('foo', 'unknown')
```

You need to make those sideload YAML files available inside the pod. And instead of copying onto the PVC it would be nice to use config maps for that.

`kiwigrid/k8s-sidecar` does exactly that: it monitors the cluster for any config map changes that have the required label (default: `kapacitor_sideload`) and will map the contents of the config maps to files inside `/var/lib/kapacitor/sideload` and removes them if the config map is deleted. Please see the https://github.com/kiwigrid/k8s-sidecar for more information. As I said before this is the exact same mechanism used be the grafana helm chart.

The reason for this change is to have better automation because it is much easier to continuously update config maps within kubernetes as to put files on the kapacitor pvc from the outside.

Hope this makes the intension behind it more clear.
Please let me know if it is still unclear.

rawkode:
Sorry @ydkn, we're happy to merge this now; but could you rebase for us, please?

ydkn:
@rawkode Thanks! Sadly the rebase removed the reviews... sorry for that.

rawkode:
Thanks @ydkn

PR #65, opened 2020-04-17
Kristinnardal2/generic oidc

kristinnardal2:
This will allow us to use generic oauth using the helm chart more
easily (this was already possible by using `envFromSecret`).

Tested with oauth from Azure.

### DIFF
--- a/charts/chronograf/values.yaml
+++ b/charts/chronograf/values.yaml
@@ -75,6 +75,9 @@ oauth:
   enabled: false
   # Used for JWT to support running multiple copies of Chronograf
   token_secret: CHANGE_ME
+  # Set if the OAuth provider implements oidc with RS256 signature
+  # use_id_token: true
+  # jwks_url: ""
   github:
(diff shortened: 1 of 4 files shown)


charts/chronograf/values.yaml
```diff
@@ -95,6 +98,20 @@ oauth:
     client_secret: CHANGE_ME
     # This is a comma separated list of Heroku organizations (OPTIONAL)
     he_orgs: ""
+  generic:
+    enabled: false
+    client_id: CHANGE_ME
```
rawkode, line 103: This should be commented out, which should cause the chart to fail if the user fails to provide the values

rawkode: This goes for all the "CHANGE_ME"'s

kristinnardal2: I commented out the "CHANGE_ME"'s, rebased the branch and squashed one of the commits.

PR #66, opened 2020-04-20
Add lifecycle hooks to Telegraf

suda:
This PR adds the ability to specify lifecycle hooks for Telegraf. This is particularly useful, when running a HA Telegraf deployment and wanting to make sure all the metrics are being flushed before Pod is destroyed.


suda:
Not sure why the lint check failed and there is not a lot of info in the logs: https://github.com/influxdata/helm-charts/pull/66/checks?check_run_id=602248453 🤔

nsteinmetz:
Error is:

```
Linting chart 'telegraf => (version: "1.7.10", path: "charts/telegraf")'
Checking chart 'telegraf => (version: "1.7.10", path: "charts/telegraf")' for a version bump...
Old chart version: 1.7.10
New chart version: 1.7.10
Error: Error linting charts: Error processing charts
```

In `telegraf/Chart.yml`, you need to increment the `version` property.

suda:
Thanks Nicolas! All green now and ready for review @rawkode :)

charts/telegraf/templates/deployment.yaml
```diff
@@ -39,6 +39,25 @@ spec:
         - name: {{ .name }}
           mountPath: {{ .mountPath }}
         {{- end }}
+        {{- if .Values.hooks }}
```
rawkode, line 42: This would lead to a failed deployment if someone has `hooks: {}` in their values.yaml.

Could we maybe use `{{- if or (.Values.hooks.preStop) (.Values.hooks.postStart) }}` ?

I've not tested this, but I think it might work :)

suda: Good call! Pushed the change 👍

Issue #45, opened 2020-03-19
Make ingress path configurable


jehie:
Hi,

Would it be possible to make the influxdb chart Ingress path configurable from the values file? At the moment the value is hard coded into "/". 

I have some use cases to change the path and also it does not work well with aws-alb-ingress-controller, which then creates an Application Load Balancer, that only forwards request from '/', so calling /query does not work.

`
{{- end }}
  rules:
  - host: {{ .Values.ingress.hostname }}
    http:
      paths:
      - path: /
        backend:
          serviceName: {{ include "influxdb.fullname" . }}
          servicePort: 8086
{{- end -}}
`

russorat:
@jehie thanks! would this be something you'd be willing to help contribute?

jehie:
Sure - I will sort the CLA and try to get this done this week :)

jgeth:
Would appreciate this for Chronograf chart as well. `/` is treated differently by different Ingress controllers. See related: https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/

PR #73, opened 2020-04-30
chore: allows for overridable ingress path

jgeth:
Modifies ingress path for chronograf and influxdb charts
Makes ingress path a template variable
Moves default path to values config
Bumps chart versions
Resolves: https://github.com/influxdata/helm-charts/issues/45

### DIFF
--- a/charts/chronograf/Chart.yaml
+++ b/charts/chronograf/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: chronograf
-version: 1.1.10
+version: 1.1.11
 appVersion: 1.8.0
 description: Open-source web application written in Go and React.js that provides
   the tools to visualize your monitoring data and easily create alerting and automation
--- a/charts/chronograf/templates/ingress.yaml
+++ b/charts/chronograf/templates/ingress.yaml
@@ -21,7 +21,7 @@ spec:
   - host: {{ .Values.ingress.hostname }}
     http:
       paths:
-      - path: /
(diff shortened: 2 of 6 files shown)


rawkode:
@jgeth Can you bump the Chart version, please?

jgeth:
> @jgeth Can you bump the Chart version, please?

@rawkode bumped versions

geoff-coppertop:
I might be a bit late to the party but when I configure ingress.path to be anything but "/" I get a 404 response, am I missing something that is needed to make this work? I'm not having any issue when I put it on a subdomain, but that doesn't match the organization in my paths that I'm looking for.

nat45928:
@geoff-coppertop I am seeing the same thing, did you find a solution?

nat45928:
I figured this out. You need to specify a re-write rule in the ingress controller.

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: influxdb
  annotations:
    kubernetes.io/ingress.class: "nginx"
    nginx.ingress.kubernetes.io/rewrite-target: /$2
spec:
  tls:
    - hosts:
      - {host}
      secretName: {host}-tls
  rules:
    - host: {host}
      http:
        paths:
          - path: /influxdb(/|$)(.*)
            pathType: Prefix
            backend:
              service:
                name: influxdb
                port:
                  number: 8086
```

See https://kubernetes.github.io/ingress-nginx/examples/rewrite/

PR #77, opened 2020-05-06
fix deprecation of ingress api version

ydkn:
The api version of `extensions/v1beta1` for the `ingress` kind is deprecated:

> warning: extensions/v1beta1/Ingress is deprecated by networking/v1beta1/Ingress and not supported by Kubernetes v1.20+ clusters.

This PR will update the api version to `networking.k8s.io/v1beta1`

### DIFF
--- a/charts/chronograf/Chart.yaml
+++ b/charts/chronograf/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: chronograf
-version: 1.1.10
+version: 1.1.11
 appVersion: 1.8.0
 description: Open-source web application written in Go and React.js that provides
   the tools to visualize your monitoring data and easily create alerting and automation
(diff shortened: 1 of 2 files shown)


rawkode:
Hi @ydkn,

We'll need to put this on hold for the time being. `networking.k8s.io/v1beta1` was added in v1.14 (IIRC).

After EKS deprecate v1.13 (June 30th, 2020), I think we can merge this without causing too many problems.

Alternatively, we could bump Chronograf chart to 2.0.0 and add a disclaimer about >= 1.14

Thoughts, @gitirabassi ?

ydkn:
Hi @rawkode 

no problem. The deprecation timeline is a valid point.

I'm just wondering why the InfluxDB chart is using the new api version then? (https://github.com/influxdata/helm-charts/blob/master/charts/influxdb/templates/ingress.yaml)

rawkode:
@ydkn That's a valid point, I wasn't aware that it was!

This was committed when the charts were still in Helm stable.

Either everybody is using 1.14 or nobody exposes the database with Ingress - I don't know which :)

Maybe we don't need to worry about 1.13 support

PR #83, opened 2020-05-07
Adding a flag to enable inputs internal plugin

### LINKED ISSUE #79: Adding input.internal
I'm trying to add the [input internal](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/internal) to the telegraf config with the following configuration:
```
## Default values.yaml for Telegraf
## This is a YAML-formatted file.
## ref: https://hub.docker.com/r/library/telegraf/tags/
...

  inputs:
    - kubernetes:
        url: "https://$HOSTIP:10250"
        bearer_token: "/var/run/secrets/kubernetes.io/serviceaccount/token"
        insecure_skip_verify: true
    - docker: 
        endpoint: "unix:///var/run/docker.sock"
        timeout: "5s"
        perdevice: true
        total: false
        docker_label_exclude:
        - "annotation.kubernetes.io/*"
    - internal:
```
But it's not added in telegraf.conf, how can I add it?

max5599:
Resolves #79

### DIFF
--- a/charts/telegraf-ds/values.yaml
+++ b/charts/telegraf-ds/values.yaml
@@ -81,3 +81,4 @@ config:
         password: ""
         user_agent: "telegraf"
         insecure_skip_verify: false
+  monitor_self: false
--- a/charts/telegraf-ds/Chart.yaml
+++ b/charts/telegraf-ds/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: telegraf-ds
-version: 1.0.8
+version: 1.0.9
 appVersion: 1.14
 deprecated: false
 description: Telegraf is an agent written in Go for collecting, processing, aggregating, and writing metrics.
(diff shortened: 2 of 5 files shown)


rawkode:
Hi @max5599,

Thank you for putting this together.

Going to suggest one change in the review, just to remove any confusion, and we'll get this merged soon after the fix.

Thank you

charts/telegraf-ds/values.yaml
```diff
@@ -81,3 +81,5 @@ config:
         password: ""
         user_agent: "telegraf"
         insecure_skip_verify: false
+  inputs:
```
rawkode, line 84: I'd like not to give any sort of indication that input configuration is available in this chart, so I'd like to suggest we rename this option. 

Current thought is `monitor_self`, but open to suggestions

max5599: No better suggestion, the change is done

PR #84, opened 2020-05-11
[influxdb] refactored backups 

monotek:
Signed-off-by: André Bauer <monotek23@gmail.com>

* refactored backup
  * uses a pvc so emptydir can't longer fill your nodes filesystem and therefore kill it
  * added configurable backup retention
  * added upload to google & azure storage


charts/influxdb/README.md
```diff
@@ -194,7 +216,7 @@ It is up to the end user to configure their own one-off restore jobs. Below is j
 apiVersion: batch/v1
 kind: Job
 metadata:
-  generateName: influxdb-restore-
+  generateName: influxdb-restore
```
naseemkullah, line 219: Please revert this.

naseemkullah:
Can you do this such that there are no breaking changes?

Also please do not reverse the file names, `backup-configmap` is fine, no need `configmap-backup`.

charts/influxdb/templates/configmap-backup.yaml
```diff
@@ -0,0 +1,37 @@
...
+    done
+
+    echo "deleting old backups"
+    find {{ .Values.backup.directory }} -not -name ".snapshot" -not -name "lost+found" -mindepth 2 -maxdepth 2 -type d -mtime +{{ .Values.backup.retentionDays }} -exec rm -r {} \;
+  upload-google.sh: |-
+    #!/bin/sh
+
+    set -ex
+
+    gcloud auth activate-service-account --key-file /var/secrets/google/{{ .Values.backup.uploadProviders.google.secretKey }}
```
naseemkullah, line 29: Please refer to inline script, key file should be optional.

naseemkullah:
Instead of creating a configmap backup, could you please make changes inline, and not change file names? This will ease reviewing. 🙏

naseemkullah:
In fact, the best would be to create a PR to just add PVC for the backup job as a disabled by default option. 🙏

charts/influxdb/README.md
naseemkullah: I think its good to have a `-` between restore and the generated id

charts/influxdb/templates/configmap-backup.yaml
monotek: Why?
Imho having it in the configmap is much cleaner as blowing up the cronjob with bash scripting.
In the end it becomes unreadable in the cronjob file.

How can the keyfile be optional?
Without it gcloud can never access the storage bucket.
The command also fails without it. See:

```
gcloud auth activate-service-account
ERROR: (gcloud.auth.activate-service-account) argument --key-file: Must be specified.
Usage: gcloud auth activate-service-account [ACCOUNT] --key-file=KEY_FILE [optional flags]
  optional flags may be  --help | --password-file | --prompt-for-password

For detailed information on this command and its flags, run:
  gcloud auth activate-service-account --help
```

naseemkullah: Sorry I was inaccurate, if you refer to the inline script, the entire gcloud auth activate command is optional, based on wether or not the key is defined. Please see https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity for more info.

While I agree not blowing up the cronjob with 3 different bash scripts is better, it is harder to detect changes when moving to a new file + making changes. Preferably we are only doing one of the two (either moving the scripts, or changing the scripts) for my tired eyes :D

monotek:
>and not change file names? 

Imho having the filenames changed makes the folder much more readable as you group the resources like statefulesets and configmaps together instead of mixing them up like now.

Also the content does not change so it does not introduce any breaking changes but makes future changes easier.

charts/influxdb/README.md
naseemkullah, line 219: Having a `-` between the prefix and the generated id is good, why remove it?

charts/influxdb/templates/configmap-backup.yaml
naseemkullah: tl;dr if we keep the bash scripts in the cornjob it will be easier for review, once the scripts are agreed upon we can move to a configmap?

charts/influxdb/README.md
monotek: Looked like a typo :D

monotek:
* added dash again
* removed gcloud auth command if no secret is set

>and not change file names? 

Imho having the filenames changed makes the folder much more readable as you group the resources like statefulesets and configmaps together instead of mixing them up like now.

Also the content does not change so it does not introduce any breaking changes but makes future changes easier.

charts/influxdb/templates/configmap-backup.yaml
naseemkullah: When possible please convert the changes inline so we can properly check the nuanced differences (e.g. another one im saying is DST_URL is replaced with bucket name, but we might want a specific folder in the bucket targetted, etc.) rather than in separate configmap.

naseemkullah:
> * added dash again
> * removed gcloud auth command if no secret is set
> 
> > and not change file names?
> 
> Imho having the filenames changed makes the folder much more readable as you group the resources like statefulesets and configmaps together instead of mixing them up like now.
> 
> Also the content does not change so it does not introduce any breaking changes but makes future changes easier.

Why would you want to have all statefulsets grouped together? I prefer that resources that are part of a component be grouped together, e.g. `backup-secret` and `backup-cronjob`...  etc.

monotek:
For me its more logical to have the same order as in kubectl.
First the resource and than the name.
But i could live with the other way around too.

naseemkullah:
> For me its more logical to have the same order as in kubectl.
> First the resource and than the name.
> But i could live with the other way around too.

I guess we are both opinionated on this. Let's get someone from influx to break the tie :p 

@rawkode could you please weigh in here? do you prefer the naming convention for templates to be:

`<component>-<type>` (`meta-statefulset`) or `<type>-<component>` (`statefulset-meta`) ?

rawkode:
I'm always of the opinion that one should group by function over type, so I would prefer:

- backup-pvc.yaml
- backup-statefulset.yaml
- backup-job.yaml

---

I'm worried about all this backup logic ending up in this chart though. I know some existed previously, and this appears to be much better and well thought out (Thank you @monotek) - but could it be deployed as it's own set of resources?

Separating the charts into `influxdb` and `influxdb-backups` seems like we could iterate on the backups with a much better velocity and not interfere with anyones database. I know this isn't par for course in Helm land, but perhaps we can start a new thing :)

monotek:
> I'm always of the opinion that one should group by function over type, so I would prefer:
> 
> * backup-pvc.yaml
> * backup-statefulset.yaml
> * backup-job.yaml
> 

ok, i'll revert the name changes.

> I'm worried about all this backup logic ending up in this chart though. I know some existed previously, and this appears to be much better and well thought out (Thank you @monotek) - but could it be deployed as it's own set of resources?
> 
> Separating the charts into `influxdb` and `influxdb-backups` seems like we could iterate on the backups with a much better velocity and not interfere with anyones database. I know this isn't par for course in Helm land, but perhaps we can start a new thing :)

Not sure what the benefit of a second chart would be? I don't think we would have much iterations anyway.

Separating it would mean that the service name and ports would need to be configured manually in the influx-backup chart instead of getting it dynamically from the influxdb chart itself.

So i would realy prefer to keep it here.

charts/influxdb/values.yaml
```diff
@@ -240,22 +239,56 @@ initScripts:
...
+    schedule: "0 2 * * *"
+    ## Number of successful finished jobs to retain
+    ##
+    historyLimit: 3
+    ## Pod annotations
+    ## ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
+    ##
+    podAnnotations: {}
+  ## Storage providers where to upload backups
+  ##
```
naseemkullah, line 269: Please bring these values one level up directly nest under `backup` instead of nesting in `cronjob`.

charts/influxdb/values.yaml
```diff
@@ -240,22 +239,56 @@ initScripts:
...
+    azure:
+      enabled: false
+      secret: ""
+      secretKey: "connection-string"
+      existingSecret: ""
+      containerName: "influxdb-container"
+      image:
+        registry: docker.io
+        repository: microsoft/azure-cli
+        tag: 2.0.24
```
naseemkullah, line 294: Please bring back 1 level up and revert the key names (gcs) as to avoid breaking changes.
```suggestion
    google:
      enabled: false
      secret: ""
      secretKey: "key.json"
      existingSecret: ""
      bucketName: "gs://bucket/influxdb"
      image:
        registry: docker.io
        repository: google/cloud-sdk
        tag: 291.0.0-alpine
    ## Azure Storage Container configuration
    ##
    azure:
      enabled: false
      secret: ""
      secretKey: "connection-string"
      existingSecret: ""
      containerName: "influxdb-container"
      image:
        registry: docker.io
        repository: microsoft/azure-cli
        tag: 2.0.24
```

charts/influxdb/templates/backup-cronjob.yaml
```diff
@@ -2,102 +2,101 @@
...
-            - |
-              influxd backup -host {{ include "influxdb.fullname" . }}.{{ .Release.Namespace }}.svc:{{ .Values.config.rpc.bind_address | default 8088 }} -portable /backup/$(date +%Y%m%d%H%M%S)
+            - name: influxdb-backup
+              image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+              env:
+                - name: INFLUXDB_ADMIN_USER_PASSWORD
+                  valueFrom:
+                    secretKeyRef:
+                      name: {{ include "influxdb.fullname" . }}-auth
+                      key: influxdb-password
```
naseemkullah, line 52: Please wrap around if statement, auth may not be set and it will look for a non-existing key.

charts/influxdb/templates/backup-configmap.yaml
```diff
@@ -0,0 +1,38 @@
...
+    gcloud auth activate-service-account --key-file /var/secrets/google/{{ .Values.backup.uploadProviders.google.secretKey }}
+    {{- end }}
+    gsutil -m rsync -r -d {{ .Values.backup.directory }}/ {{ .Values.backup.uploadProviders.google.bucketName }}
+  upload-azure.sh: |-
+    #!/bin/sh
+
+    set -ex
+
+    az storage blob sync --source {{ .Values.backup.directory }} --container {{ .Values.backup.uploadProviders.azure.containerName }}
+{{ end }}
```
naseemkullah, line 38: Can you please put these back in line in the cronjob to more easily compare any changes in the scripts?

charts/influxdb/README.md
```diff
@@ -98,12 +98,28 @@ The command removes all the Kubernetes components associated with the chart and
...
-| backup.azure | Azure Blob Storage config | `nil`
-| backup.schedule | Schedule to run jobs in cron format | `0 0 * * *` |
-| backup.annotations | Annotations for backup cronjob | {} |
-| backup.podAnnotations | Annotations for backup cronjob pods | {} |
+| `backup.enabled`                                  | enable InfluxDB backup                                                                                                                                                                    | `false`                                                 |
+| `backup.directory`                                | directory where backups are stored in                                                                                                                                                     | `"/backups"`                                            |
+| `backup.retentionDays`                            | retention time in days for backups (older backups are deleted)                                                                                                                            | `10`                                                    |
```
naseemkullah, line 103: Please specify that this is for PVC and not the long term object store.

Do you expect some users to just keep backups in a PVC and not in an object store?

charts/influxdb/values.yaml
monotek: why?

charts/influxdb/values.yaml
monotek: i've updated major chart version and added update docu to the readme so no problem to introduce some changes.

charts/influxdb/templates/backup-configmap.yaml
monotek: as dicsussed earlier i'd like to keep it that way as it makes everything much more readable.

charts/influxdb/README.md
monotek: some people may not even have an object store available.
also the retention works on the object store by syncing this dir.

naseemkullah: the object store retention should be managed independently such as via https://cloud.google.com/storage/docs/lifecycle

charts/influxdb/templates/backup-configmap.yaml
naseemkullah: But you are also changing the nature of the scripts, it is more burdensome for review

naseemkullah: Very well, but it would be easier to review it if iterating on the following separately:
- extract the scripts
- change the scripts

charts/influxdb/values.yaml
naseemkullah: These are breaking changes that are purely stylistic though, so not worth it imho

charts/influxdb/values.yaml
naseemkullah: The same reason we don't have a `statefulset:` value, whatever that is :)

naseemkullah: also needless breaking changes

charts/influxdb/README.md
naseemkullah: hence the request to specify it for clarity

charts/influxdb/values.yaml
monotek: done

charts/influxdb/README.md
monotek: done

charts/influxdb/templates/backup-cronjob.yaml
monotek: Without this the backup script won't work as it queries the existing databases via this user.
I've added a note to the values file that you need the defaultuser.

charts/influxdb/values.yaml
monotek: But even with 1 level up the changes are breaking.
Imho the renaming is a improvement to the current one as they are much more consistent over the storageproviders.

charts/influxdb/templates/backup-configmap.yaml
naseemkullah: Pasting scripts here for easier comparison....


```sh
backup.sh: |-
    #!/bin/bash
    set -e
    DATE="$(date +%Y%m%d_%H%M%S)"
    for DATABASE in $(influx -username {{ .Values.setDefaultUser.user.username }} -password "${INFLUXDB_ADMIN_USER_PASSWORD}" -host {{ include "influxdb.fullname" . }}.{{ .Release.Namespace }}.svc -port {{ .Values.config.http.bind_address | default 8086 }} -execute 'SHOW DATABASES' | sed -e '1,3d'); do
        echo "backuping ${DATABASE} db to {{ .Values.backup.directory }}/${DATABASE}"
        mkdir -p {{ .Values.backup.directory }}/${DATABASE}
        influxd backup -host {{ include "influxdb.fullname" . }}.{{ .Release.Namespace }}.svc:{{ .Values.config.rpc.bind_address | default 8088 }} -portable -db ${DATABASE} {{ .Values.backup.directory }}/"${DATABASE}/${DATE}"
    done
```
vs
```sh
              influxd backup -host {{ include "influxdb.fullname" . }}.{{ .Release.Namespace }}.svc:{{ .Values.config.rpc.bind_address | default 8088 }} -portable /backup/$(date +%Y%m%d%H%M%S)
```

```
    echo "deleting old backups"
    find {{ .Values.backup.directory }} -not -name ".snapshot" -not -name "lost+found" -mindepth 2 -maxdepth 2 -type d -mtime +{{ .Values.backup.retentionDays }} -exec rm -r {} \;
```

vs not applicable, since backup file is transitory in emtpyDir in all cases

  upload-google.sh: |-
```
    set -ex
    {{ if .Values.backup.uploadProviders.google.secret }}
    gcloud auth activate-service-account --key-file /var/secrets/google/{{ .Values.backup.uploadProviders.google.secretKey }}
    {{- end }}
    gsutil -m rsync -r -d {{ .Values.backup.directory }}/ {{ .Values.backup.uploadProviders.google.bucketName }}
```
vs
```sh
              if [ -n "$KEY_FILE" ]; then
                gcloud auth activate-service-account --key-file $KEY_FILE
              fi
              gsutil -m cp -r /backup/* "$DST_URL"
```

  upload-azure.sh: |-

```sh
    #!/bin/sh
    set -ex
    az storage blob sync --source {{ .Values.backup.directory }} --container {{ .Values.backup.uploadProviders.azure.containerName }}
```

vs

```sh
              az storage container create --name "$DST_CONTAINER"
              az storage blob upload-batch --destination "$DST_CONTAINER" --destination-path "$DST_PATH" --source "$SRC_URL"
```

monotek: So you still want separate pr's? 

Imho its not that hard to review.
You have an entirely new file here.
The file, the bash stuff was taken from, is missing some lines now.

Still think its a waste of time but if you insist i'll do it...

charts/influxdb/templates/backup-configmap.yaml
```diff
@@ -0,0 +1,38 @@
...
+    {{- include "influxdb.labels" . | nindent 4 }}
+data:
+  backup.sh: |-
+    #!/bin/bash
+
+    set -e
+
+    DATE="$(date +%Y%m%d_%H%M%S)"
+
+    for DATABASE in $(influx -username {{ .Values.setDefaultUser.user.username }} -password "${INFLUXDB_ADMIN_USER_PASSWORD}" -host {{ include "influxdb.fullname" . }}.{{ .Release.Namespace }}.svc -port {{ .Values.config.http.bind_address | default 8086 }} -execute 'SHOW DATABASES' | sed -e '1,3d'); do
```
naseemkullah, line 16: if you are still backing up all dbs, why just not specify any as in the original script?

charts/influxdb/values.yaml
naseemkullah: In which case is it more consistent?

bucketname was intended to be destination url as to specify a "folder" in the bucket, why change this?

imho these stylistic changes are not worth breaking the chart for users using this in production today.

charts/influxdb/templates/backup-cronjob.yaml
naseemkullah: What happens with those queries if the user does not exist and the dbs are not user/pass protected? If it breaks define the args based on if 

While it is not a best practice, it is surely possible to have passwordless dbs and those users should be able to backup their data. https://github.com/influxdata/helm-charts/blob/e020a4b449041c31449d867e12e772e0224f9dd4/charts/influxdb/values.yaml#L99-L100 as this is set by default, backups should adhere to that as well.

charts/influxdb/templates/backup-configmap.yaml
naseemkullah: Well I pasted the contents here to help me, but yeah it would make review much easier imho

charts/influxdb/templates/backup-configmap.yaml
monotek: To have the backups in separate files. Makes restore easier...

charts/influxdb/templates/backup-configmap.yaml
monotek: Tomorrow....

charts/influxdb/README.md
```diff
@@ -215,8 +231,8 @@ spec:
             - |
               gsutil -m cp -r gs://<PATH TO BACKUP FOLDER>/* /backup
           volumeMounts:
-            - name: backup
-              mountPath: /backup
+            - name: {{ include "influxdb.fullname" . }}-backups
+              mountPath: {{ .Values.backup.directory | quote }}
```
naseemkullah, line 235: this is not a helm template, it is an example raw yaml of a restore job, please revert

charts/influxdb/README.md
```diff
@@ -228,8 +244,8 @@ spec:
         - name: influxdb-restore
           image: influxdb:1.7-alpine
           volumeMounts:
-            - name: backup
-              mountPath: /backup
+            - name: {{ include "influxdb.fullname" . }}-backups
+              mountPath: {{ .Values.backup.directory | quote }}
```
naseemkullah, line 248: this is not a helm template, it is an example raw yaml of a restore job, please revert

charts/influxdb/templates/backup-configmap.yaml
naseemkullah: How does it make it easier? Here is an example restore job, please see the script to restore from the 1 backup folder:

https://github.com/influxdata/helm-charts/blob/master/charts/influxdb/README.md#restores

```sh
              for db in $(influx -host $INFLUXDB_HOST -execute 'SHOW DATABASES' | tail -n +5); do
                influxd restore -host $INFLUXDB_HOST:8088 -portable -db "$db" -newdb "$db"_bak /backup
              done
```

It is a little rough around the edges but what would the new restore procedure look like? Furthermore would it make restoring the entire db easier as well?

charts/influxdb/templates/backup-configmap.yaml
naseemkullah: No rush... Thanks 🙏 😅

monotek:
Nevermind. We'll extract the backup function in a new chart....

rawkode:
Hi @monotek,

Thank you for your PR and trying to improve our backup process. I know it's been difficult, but this is a chart with long term users that rely on this for their production monitoring and @naseemkullah just has their best interests in mind.

I'd like to discuss how we can iterate in this while not breaking any existing users.

If you're happy to do so, we can work together.

Thanks again.

naseemkullah:
Iteratively is the key word here, if we were to extract it, we should extract it as is, and iterate over it after, or vice versa. Would not be ideal to have duplicated efforts (backups within the chart going one direction and extracted backup chart taking another approach).

naseemkullah:
Also it may be of interest to open an issue to discuss and pile up stylistic breaking changes, and implement them when a technical breaking change is required.

naseemkullah:
With that said @monotek had excellent suggestions that could be introduced in a non breaking manner (most notably add option to use PVC for backups and enable user-auth if exists for backup script).

PR #88, opened 2020-05-19
feat: add ability to provide volumes and mountPoints

rawkode:
Previously, support was added for this with the `volumeMounts` key. I've opted not to modify this and instead provide a more flexible solution through `volumes` and `mountPoints` keys.

We'll aim to deprecate the older method in a future release.


rawkode:
Build is failing because Quay.io is down.

CC @nsteinmetz  @rajasree-1993 @Jim-Walk

claymcleod:
Should probably update the README as well 👍

PR #94, opened 2020-05-23
feat: add option to use PVC for backups -- thanks monotek

naseemkullah:
cc @monotek

Heavily inspired by @monoteks suggestions, this allows users to use a PVC either as their long term or intermediary storage for backups. Disabled by default.

Anyway we could get your approval @monotek?

### DIFF
--- a/charts/influxdb/Chart.yaml
+++ b/charts/influxdb/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: influxdb
-version: 4.4.11
+version: 4.5.0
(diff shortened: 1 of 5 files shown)


naseemkullah:
> I would vote for "Enabled by default" as emptydir could bring your node down if the backup fills your disk completly.

Good point, imho a nice default is to set an ephemeral storage request to the job. I've added the ability to provide resources to the job pod, setting a storage request by default. 

I've also added a note in the docs.

naseemkullah:
bump

PR #100, opened 2020-05-28
Fix rendering of input configuration.

### LINKED ISSUE #97: telegraf chart: unable to specify input tags
I'm unable to specify tags on my inputs using the `telegraf` chart.

In my values.yml I have something like this:

```
  inputs:
  - prometheus:
      kubernetes_services: 
      - http://blackbox-exporter:9115/probe?....
      metric_version: 2
      tags:
        foo: bar
```

Which gets templated into the following TOML:

```
[[inputs.prometheus]]
  alias = "blackbox"
  kubernetes_services = [
    "http://blackbox-exporter:9115/probe?...
  ]
  metric_version = 2
  [[inputs.prometheus.tags]]
    foo = "bar"
```

However, `tags` is a table, not an array, so I think it should be `[inputs.prometheus.tags]` instead of `[[inputs.prometheus.tags]]`.  Am I doing something wrong or is the template incorrect?  Any workarounds?

https://github.com/influxdata/helm-charts/blob/4eb7881ddbe9604cfb40cbad99c320e318baf0c6/charts/telegraf/templates/_helpers.tpl#L238

ambercs:
Nested TOML tables should only be surrounded by a single bracket.

Fixes #97.

Co-authored-by: Zac Bergquist <bergquistz@vmware.com>

### DIFF
--- a/charts/telegraf/Chart.yaml
+++ b/charts/telegraf/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: telegraf
-version: 1.7.14
+version: 1.7.16
 appVersion: 1.14
 deprecated: false
 description: Telegraf is an agent written in Go for collecting, processing, aggregating, and writing metrics.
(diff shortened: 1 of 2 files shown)


rawkode:
@ambercs The version will need bumped for the tests to pass :+1:

ambercs:
@rawkode Hello, the version has been bumped, please let me know if there is anything else that I need to do.

PR #112, opened 2020-06-25
Typo in template

cocampbe:
reference to the value is out of order. Supposed to be backup.persistence.


naseemkullah:
Thanks @cocampbe

PR #116, opened 2020-06-29
remote backups

cocampbe:
While trying to setup the backup cronjob ran into an issue with container spec. I don't hace gcs or azure to do remote offsite copies. Being on prem the backup would be on NFS and backed up to VTL onsite. This patch allows for use cases where remote backup is not available.

### DIFF
--- a/charts/influxdb/templates/backup-pvc.yaml
+++ b/charts/influxdb/templates/backup-pvc.yaml
@@ -12,7 +12,7 @@ spec:
     requests:
       storage: {{ .Values.backup.persistence.size | quote }}
 {{- if .Values.backup.persistence.storageClass }}
-{{- if (eq "-" .Values.persistence.backup.storageClass) }}
+{{- if (eq "-" .Values.backup.persistence.storageClass) }}
   storageClassName: ""
 {{- else }}
   storageClassName: "{{ .Values.persistence.backup.storageClass }}"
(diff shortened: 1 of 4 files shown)


charts/influxdb/templates/backup-pvc.yaml
```diff
@@ -12,10 +12,10 @@ spec:
     requests:
       storage: {{ .Values.backup.persistence.size | quote }}
 {{- if .Values.backup.persistence.storageClass }}
-{{- if (eq "-" .Values.persistence.backup.storageClass) }}
+{{- if (eq "-" .Values.backup.persistence.storageClass) }}
```
naseemkullah, line 15: 👍 

Please keep this change.

charts/influxdb/templates/backup-pvc.yaml
```diff
@@ -12,10 +12,10 @@ spec:
...
   storageClassName: ""
 {{- else }}
-  storageClassName: "{{ .Values.persistence.backup.storageClass }}"
+  storageClassName: "{{ .Values.backup.persistence.storageClass }}"
```
naseemkullah, line 18: 👍  and this one.

naseemkullah:
Hi @cocampbe I've just noticed this in https://github.com/influxdata/helm-charts/pull/112 which has just been merged, please let us know if the backups to persistent volume work as intended now! Thanks.

Issue #111, opened 2020-06-18
[telegraf] No support for `[[inputs.postgresql_extensible.query]]`?


florianrusch:
Hi everybody,

currenty I'm migrating my local telegraf docker instance to a helm instance.
I'm currently using the input plugin `postgresql_extensible` with which I can define some queries in this format:

```toml
[[inputs.postgresql_extensible]]
  address = "..."

  [[inputs.postgresql_extensible.query]]
    script="/app/foobar.sql"
    version="1200"
    withdbname=false
    tagvalue="tenant"
    measurement="something"
```

This is my current yaml version of it:

```yaml
config:
  - inputs:
    - postgresql_extensible:
      address: "..."
      query:
        script: "/app/foobar.sql"
        version: "1200"
        withdbname: "false"
        tagvalue: "tenant"
        measurement: "something"
```

If I try to run this I'm getting this error message:

```
2020-06-18T12:10:31Z I! Starting Telegraf 1.14.4
2020-06-18T12:10:31Z I! Using config file: /etc/telegraf/telegraf.conf
2020-06-18T12:10:31Z E! [telegraf] Error running agent: Error parsing /etc/telegraf/telegraf.conf, line 75: cannot unmarshal TOML table into postgresql_extensible.query (need struct or map)
```

This is the produced TOML at that possition:
```toml
[[inputs.postgresql_extensible]]
  address = "..."
  [inputs.postgresql_extensible.query]
    measurement = "something"
    script = "/app/foobar.sql"
    tagvalue = "tenant"
    version = "1200"
    withdbname = "false"
```

As you can see there is just one bracket pair (`[...query]`) and not two (`[[...query]]`).
I've looked into the `_helpers.tpl` and for me it looks like, that this use case isn't support currently.

Am I correct and does someone know a workaround or something?

florianrusch:
Or is this a bug??

For the outputs there are two bracket pairs:
https://github.com/influxdata/helm-charts/blob/c0904994d655da1c78c800d5ddcb7feb578728c5/charts/telegraf/templates/_helpers.tpl#L137

And aso for the proccessors:
https://github.com/influxdata/helm-charts/blob/c0904994d655da1c78c800d5ddcb7feb578728c5/charts/telegraf/templates/_helpers.tpl#L321

But not for the input plugins:
https://github.com/influxdata/helm-charts/blob/c0904994d655da1c78c800d5ddcb7feb578728c5/charts/telegraf/templates/_helpers.tpl#L232

[...]

PR #117, opened 2020-06-29
WIP: feat: add possibility to use separately configmap

florianrusch:
Regarding #111 I've implemented a way to provide a TOML configuration without writing complex YAML stuff.

I've tested it with theses commands:

```shell
# There should be the normal configmap like there were no changes:
helm upgrade --install --debug --dry-run my-release .

# The normal configmap which normally will be provided by the helm chart should
# be missing and at the deployment this should be existing
# `spec.template.spec.volumes[0].configMap.name: telegraf-configuration`
helm upgrade --install --debug --dry-run my-release \
    --set existingConfigMapName=telegraf-configuration .
```

---
- [ ] CHANGELOG.md updated
- [ ] Rebased/mergable
- [ ] Tests pass

### DIFF
--- a/charts/telegraf/Chart.yaml
+++ b/charts/telegraf/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: telegraf
-version: 1.7.17
+version: 1.7.18
 appVersion: 1.14
 deprecated: false
 description: Telegraf is an agent written in Go for collecting, processing, aggregating, and writing metrics.
--- a/charts/telegraf/README.md
+++ b/charts/telegraf/README.md
@@ -64,3 +64,77 @@ To configure the chart, do either of the following:
   ```
 
   > **Tip**: Use the default [values.yaml](values.yaml).
+
+### Use a custom Telegraf configuration
+
+You can provide the Telegraf configuration as YAML-Value on the chart directly or by provisioning a separate ConfigMap.
+
+### 1. Provide the Configuration as YAML directly
+
+For example:
+
+```yaml
+# Content of example-values.yaml
+config:
+  agent:
+    interval: "10s"
+    round_interval: true
+    metric_batch_size: 1000
+    metric_buffer_limit: 10000
+    collection_jitter: "0s"
+    flush_interval: "10s"
+    flush_jitter: "0s"
+    precision: ""
+    debug: false
+    quiet: false
+    logfile: ""
+    hostname: "$HOSTNAME"
+    omit_hostname: false
+  processors:
+    - enum:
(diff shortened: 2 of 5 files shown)


pravindamodaran:
@florianrusch Is it possible to specify configmaps similarly for influxdb, kapacitor and chronograf too?

zmb3:
My opinion is that helm should still own the config map, and the config should just be a multi-line TOML string in values.yaml.  This eliminates the complexity and error cases arising from trying to use templates to convert YAML to TOML, rather than simply providing an escape hatch.

florianrusch:
@zmb3 that's a great idea! I will try to implement this in the course of the day.👍

rawkode:
I'm not really keen on the YAML to TOML, we've had nothing but problems with this on the charts. I think allowing straight up TOML is much better and easier for our users to copy from existing online examples.

rawkode:
Exactly as @zmb3 said 😀

pravindamodaran:
> @pravindamodaran I'm not very familiar with the other helm charts. But it should be possible to adopt the changes, regardless of which pattern has been implemented.

afaik, its difficult to specify custom configuration for other TICK charts too (either as env vars, or through values.yaml, which again defeats the purpose). It will be really useful to specify the toml config directly. I'll try to adopt the changes you make for telegraf to other charts too and see how it works out

florianrusch:
Sorry for the delay. I haven't been feeling too well the last two days.

I'm thinking about to provide all three options for configuring telegraf - so the user can decide on his use case. So I would like to provide this three options:

1. External ConfigMap – this is what I have already implemented. When this value is set, it disables the other two options.
2. TOML config – Here the user can provide a multiline string with the Toml configuration directly in the helm values. When this value is set, it disables the third option.
3. YAML config – The current behaviour with the yaml stuff. (Maybe we should deprecate this?)

I'd like to standardize the names - but that might lead to breaking changes :-/
What do you think about this?

1. `existingTomlConfigMap`
2. `tomlConfig`
3. `yamlConfig`

I'm not really sure how to handle renaming the current `config' value, as this would be a breaking change. Or should we leave it at `config' to minimize user confusion?

Mr-Howard-Roark:
Doesn't this fail to solve the problem where the service logic that creates corresponding listeners still depends on the config inputs?  https://github.com/influxdata/helm-charts/blob/master/charts/telegraf/templates/service.yaml#L18-L19

ie- if you only include an existing configmap, no corresponding listeners will be created for the service.  So, for this to work in its existing state, I think you need to include both the config in the values file AND the existing configmap

PR #125, opened 2020-07-12
set collectd service as a UDP service

### LINKED ISSUE #107: The collectd service uses TCP when it should use UDP
The [collectd service template](https://github.com/influxdata/helm-charts/blob/master/charts/influxdb/templates/service.yaml#L28) does not specify the UDP protocol and therefore launches as TCP. However, the [collectd service uses UDP](https://docs.influxdata.com/influxdb/v1.8/supported_protocols/collectd/) and not TCP. Manually editing the deployment to use UDP allowed traffic to flow in.

aureq:
This commit sets `collectd` service as UDP instead of TCP.

fixes #107

### DIFF
--- a/charts/influxdb/Chart.yaml
+++ b/charts/influxdb/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: influxdb
-version: 4.7.1
+version: 4.7.2
 appVersion: 1.8.0
 description: Scalable datastore for metrics, events, and real-time analytics.
 keywords:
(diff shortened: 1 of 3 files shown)


naseemkullah:
Thanks @aureq! Please bump the chart version.

naseemkullah:
We just released 4.7.1 https://github.com/influxdata/helm-charts/blob/master/charts/influxdb/Chart.yaml can you please rebase and bump to 4.7.2?

aureq:
Hi @naseemkullah Updated the PR. Let me know if this is good on your side.

naseemkullah:
lgtm @aureq but we'll need a second approval

PR #129, opened 2020-07-16
Mount license from secret using configurable key, instead of hardcoded key

blai:
As per documented configuration in [values.yaml](https://github.com/influxdata/helm-charts/blob/master/charts/influxdb-enterprise/values.yaml#L17), when `license` is using `secret` configuration, `key` can be set to anything.

Without this patch, `/var/run/secrets/influxdb/license.json` will be mounted as a directory!! Making the bug very difficult to discover.

This patch enables the configuration work as per documentation.

### DIFF
--- a/charts/influxdb-enterprise/templates/data-statefulset.yaml
+++ b/charts/influxdb-enterprise/templates/data-statefulset.yaml
@@ -40,6 +40,9 @@ spec:
       - name: license
         secret:
           secretName: {{ .Values.license.secret.name }}
+          items:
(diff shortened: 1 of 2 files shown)


blai:
I tried `subPath: {{ .Values.license.secret.key }}`, it didn't work for me @rawkode

rawkode:
@blai We merged your change. I will work out what happened with the `subPath`, as I actually recall having a similiar issue with the TLS changes too and had to use `items`.

I'll raise with the Kubernetes org when I've worked it out

PR #133, opened 2020-07-17
Fix underscores and dashes in service.yaml

giannoug:
This PR fixes service.yaml for `tcp_listener` and `udp_listener`.

When you set `udp_listener` or `tcp_listener` in `values.yaml`, the comparison fails and the service isn't correctly setup. Fixing the comparison shows that the Kubernetes port name is also incorrect, it needs dashes instead of underscores. This PR fixes both issues for each listener.

- [x] CHANGELOG.md updated
- [x] Rebased/mergable
- [ ] Tests pass

### DIFF
--- a/charts/telegraf/Chart.yaml
+++ b/charts/telegraf/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: telegraf
-version: 1.7.19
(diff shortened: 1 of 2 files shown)


timhallinflux:
chart version needs to be bumped

Issue #127, opened 2020-07-15
[influxdb-enterprise] config reference error


chobbs:
Misalignment of how the values are [expected to be defined](https://github.com/influxdata/helm-charts/blob/master/charts/influxdb-enterprise/values.yaml#L157-L158) and how they are [referenced](https://github.com/influxdata/helm-charts/blob/master/charts/influxdb-enterprise/templates/data-statefulset.yaml#L53-L55) (also for meta, expected config and referenced in chart). Changing the references to 
`.Values.[meta|data].https.secret.key` appears to have fix the issue.

rawkode:
@chobbs do you want to submit a PR?

chobbs:
PR submitted - https://github.com/influxdata/helm-charts/pull/135

PR #135, opened 2020-07-21
fix(influxdb-enterprise): key references

chobbs:
Update key references to align with the expected values configured in the chart


rawkode:
Closes #127

timhallinflux:
does this need a chart version bump?

rawkode:
Yeah, it does.

chobbs:
reset

PR #149, opened 2020-07-31
feat(influxdb): add backup retention

jakubkrzywda:
This PR enables automatic time-based retention of InfluxDB backups with implementation for S3 and placeholders for other storage providers.

- [ ] CHANGELOG.md updated
- [ ] Rebased/mergable
- [ ] Tests pass

### DIFF
--- a/charts/influxdb/Chart.yaml
+++ b/charts/influxdb/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: influxdb
-version: 4.8.2
+version: 4.8.9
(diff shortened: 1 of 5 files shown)


rawkode:
Hi.

I'm a little worried about adding too much responsibility to this chart. I see that you've provided an implementation for S3, but this will fail for Azure and GCP.

With regards to S3, I'd generally recommend object expiration; but I appreciate that you also check that the user will be left with at least N backups.

I need to think about this a little bit. Do you feel strongly that this chart should be responsible for this task?

PR #153, opened 2020-08-10
Support being able to set securityContext

michaelandrepearce:


rawkode:
Please bump the chart version 👍

charts/telegraf/templates/deployment.yaml
```diff
@@ -21,6 +21,10 @@ spec:
 {{ toYaml .Values.podAnnotations | indent 8 }}
         {{- end }}
     spec:
+{{- if .Values.securityContext }}
+      securityContext:
+{{ toYaml .Values.securityContext | indent 8 }}
+{{- end }}
```
naseemkullah, line 27: Please see and replicate https://github.com/helm/helm/blob/273d0364be30e3bbaa73041dceea2a338f666abe/pkg/chartutil/create.go#L269-L273 we should have both `podSecurityContext` and `securityContext` values

michaelpearce-gain: to be honest i actually took from https://github.com/influxdata/helm-charts/blob/master/charts/influxdb/templates/statefulset.yaml if you can show me there where it is there? I cannot see it, I was mearly aligning with capability avaialble in the influxdb chart,

michaelandrepearce:
@rawkode bumped the chart version

charts/telegraf/templates/deployment.yaml
naseemkullah: touché

naseemkullah: mind you we wont be able to add without breaking because securityContext would go from meaning pod security context to container security context

naseemkullah: anyways not a blocker

timhallinflux (dismissed):

rawkode (dismissed):

rawkode:
@timhallinflux What happened ? :)

timhallinflux:
Not sure.  the changes were made...but the merge was blocked.  So, I've forced the merge.

PR #155, opened 2020-08-11
chore: remove verbose curl logging

### LINKED ISSUE #132: [influxdb-enterprise] Add support TLS log debug configuration
With TLS certs configured, the logs become very verbose, please add configuration to allow turning on/off the TLS debug log in data pod  for [meta}data] pods, do the same)

rawkode:
Closes #132


chobbs:
based on the customer  - i'm sure the ask here will be to parameterize the logging for  both for the curl register and the influxd logs -- any thoughts on maybe passing a flag that turns them both on silent?

PR #173, opened 2020-08-27
Version Bumps

rawkode:
There was an InfluxDB Enterprise PR merged 6 days ago that caused a few charts to publish, as its version wasn't incremented.

This should hopefully get everything up to date.

### DIFF
--- a/charts/influxdb/Chart.yaml
+++ b/charts/influxdb/Chart.yaml
@@ -1,23 +1,23 @@
 apiVersion: v1
 name: influxdb
-version: 4.8.3
+version: 4.8.4
 appVersion: 1.8.0
 description: Scalable datastore for metrics, events, and real-time analytics.
 keywords:
-- influxdb
-- database
-- timeseries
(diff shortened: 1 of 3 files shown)


charts/influxdb/Chart.yaml
```diff
@@ -1,23 +1,23 @@
...
-- name: rawkode
-  email: rawkode@influxdata.com
-- name: gitirabassi
-  email: giacomo@influxdata.com
-- name: aisuko
-  email: urakiny@gmail.com
-- name: naseemkullah
-  email: naseem@transit.app
+  - name: rawkode
+    email: rawkode@influxdata.com
```
timhallinflux, line 16: I think you want to change this ?  And add @chobbs

chobbs: I will be adding to the charts today @timhallinflux

rawkode:
With that in mind, let's merge and fixup maintainers next?

PR #179, opened 2020-09-02
chore:bump version

chobbs:


rawkode (changes requested):
I don't think you should bump appVersion, not unless InfluxDB 1.9 is out 🙂

rawkode (approved):

PR #188, opened 2020-09-09
[telegraf-ds] Make inputs configurable

mkerins:
Modifying `config.inputs` allows the addition but not removal of predefined inputs. To disable an input the `ConfigMap` must be modified after the chart is installed. For example, the `docker` input will fail if using a different container runtime. It would be better if the list of enabled inputs could be configured in `values.yaml` in the same way outputs are configured.

### DIFF
--- a/charts/telegraf-ds/Chart.yaml
+++ b/charts/telegraf-ds/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: telegraf-ds
-version: 1.0.16
+version: 1.0.17
 appVersion: 1.14
 deprecated: false
 description: Telegraf is an agent written in Go for collecting, processing, aggregating, and writing metrics.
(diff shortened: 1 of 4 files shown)


mkerins:
Need to follow corporate PR policy.

nsteinmetz:
It was a choice to make the inputs static to have an "opinionated chart".

mer2github:
@nsteinmetz With K8s 1.20 on Azure AKS moving to containerd by default, there is no way to configure the chart to avoid a lot of unhelpful error messages muddying the log. I respect the chart's opinion on always enabling the docker input plug-in, but now disagree.

nsteinmetz:
@mer2github I no longer have time to maintain this chart. Feel free to open a PR - the issue I think is that there is no containerd input plugin yet (if any)

As things change, it's normal that helm chart may change over time ;-)

PR #189, opened 2020-09-10
prevent random password change on upgrade

trombka:
Currently, the passwords generated in `*-influxdb2-auth` secret are changed on every upgrade. However, the job setting up the admin account is executed only after installation. This makes the secret inconsistent with the running configuration and causes confusion.
This is a known issue when using `randAlphaNum` discussed [here](https://github.com/helm/charts/issues/5167). I have applied the solution suggested in a comment there.
- [ ] CHANGELOG.md updated
- [X] Rebased/mergable
- [ ] Tests pass
- [X] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

### DIFF
--- a/charts/influxdb2/Chart.yaml
+++ b/charts/influxdb2/Chart.yaml
@@ -5,7 +5,7 @@ name: influxdb2
 description: A Helm chart for InfluxDB v2
 home: https://www.influxdata.com/products/influxdb-overview/influxdb-2-0/
 type: application
-version: 1.0.6
+version: 1.0.7
 maintainers:
 - name: rawkode
   email: rawkode@influxdata.com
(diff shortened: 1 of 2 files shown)


trombka:
It seems you use an older version of Helm in the CI without the `lookup` function. At least for linting.

rawkode:
@trombka is the secret updated even when using `--reuse-values` ? 😲

trombka:
> @trombka is the secret updated even when using `--reuse-values` ? astonished

Yes.
It has little to do with resuing chart values since it's an empty password or token value that triggers random value generation. Just try this several times:
```bash
helm upgrade --install --reuse-values test influxdata/influxdb2; kubectl get secret test-influxdb2-auth -o jsonpath='{.data.admin-password}' | base64 -d; echo;
```
Every time you get a different value, but only the first one generated is set in the InfluxDd instance.

rawkode:
Ok. I'll see how to get the lookup function available during our testing process. Thank you

brsolomon-deloitte:
This is a serious bug triggered by something simple like deploying chronograf with a secret ref pointing to influx-auth.

Screen Shot 2022-02-17 at 2 12 57 PM

What is holding up this PR @rawkode @trombka ?

alespour:
To use existing secret, set its name as `adminUser.existingSecret` value. Added in PR #279.

PR #191, opened 2020-09-14
[telegraf-operator] Keep support for helmv2

### LINKED ISSUE #187: [telegraf-operator] No support for helm2 (but declared)
Telegraf-operator helm charts declares support for `Helm v2 or later`, but `apiVersion: v2` is supported only by helm3. Either `Readme.md` or `Chart.yaml` should be updated

rel:
 
* https://helm.sh/docs/topics/charts/#the-apiversion-field
* https://github.com/influxdata/helm-charts/tree/master/charts/telegraf-operator#prerequisites

sumo-drosiek:
Keep support for helm2 as it's not end of life yet

Fixes #187

### DIFF
--- a/charts/telegraf-operator/Chart.yaml
+++ b/charts/telegraf-operator/Chart.yaml
@@ -1,4 +1,4 @@
-apiVersion: v2
+apiVersion: v1
 name: telegraf-operator
 description: A Helm chart for Kubernetes to deploy telegraf-operator
 keywords:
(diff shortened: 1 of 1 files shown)


russorat:
@rawkode i can't recall if there was a reason we did not want to support helm2 or not. any ideas?

rawkode:
@russorat it's not that this chart removed support for Helm 2, it's that we opted not to support Helm 2 for any new charts we wrote. 

InfluxDB 2 and this operator never supported Helm 2.

I don't really see any reason to support Helm 2 either, as existing workloads on Helm 2 can coexist with workloads deployed with Helm 3.

Can you elaborate on your reasoning here, @sumo-drosiek?

Also, there's already automation in the root of the repo to build and publish the charts 👍

sumo-drosiek:
@rawkode the reason for this PR, is that we want to use telegraf-operator as our chart dependency and we didn't drop the support for the helm2 yet.

We already find the workaround for this, so I will close that PR and prepare fix for README.md

sumo-drosiek:
Sorry, accidentally pushed our local changes 🤦

PR #207, opened 2020-10-08
chore: improve release process to be more forgiving

rawkode:
I'm hoping that this means we only trigger a release
when the push is to the master branch and a Chart.yaml was
modified.

We've found that automatically releasing on push to master alone
causes a lot of problems, especially as we forget to encourage contributors
to bump the version; even reminding them isn't a great experience.

This should allow us to have a successful build without such a change
and then manually update the Chart version ourselves, without build
failures

### DIFF
--- a/.github/workflows/helm-charts-release.yaml
+++ b/.github/workflows/helm-charts-release.yaml
@@ -1,9 +1,22 @@
 name: "helm-charts/release"
 
+# I'm hoping that this means we only trigger a release
+# when the push is to the master branch and a Chart.yaml was
+# modified.
+#
+# We've found that automatically releasing on push to master alone
+# causes a lot of problems, especially as we forget to encourage contributors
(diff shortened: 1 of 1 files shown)


naseemkullah:
lint on PR is supposed to catch this 🤔 
How about adding Chart version bumped as a check in the PR template?

sumo-drosiek:
@naseemkullah PR check won't work in case of two concurrent PRs bump the version independently

rawkode:
@naseemkullah Also, I don't want this to be a blocker for contributors. I'd rather I had to increase the version number for them.

Time and time again, sadly, this has caused failed builds, failed releases, and a pretty horrible experience for people contributing.

I'm keen to just remove it

naseemkullah (approved):

sumo-drosiek (approved):

PR #212, opened 2020-10-09
Askainet influxdb2 ingress

rawkode:


timhallinflux (approved):

charts/influxdb2/templates/service.yaml
```diff
@@ -1,16 +1,13 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ template "influxdb.fullname" . }}
-  labels:
-    {{- include "influxdb.labels" . | nindent 4 }}
-    app.kubernetes.io/component: backup
+  name: { { template "influxdb.fullname" . } }
+  labels: { { - include "influxdb.labels" . | nindent 4 } }
```
sumo-drosiek, line 5: It doesn't seem to be correct helm template

sumo-drosiek: Same for code below

PR #223, opened 2020-10-22
fix: bump resource limits for init job

pdrastil:
Resource limits on init job might fail deployment of InfluxDB cluster. Main offender is hardcoded 10MB RAM limit on final container which is too tight and in some circumstances results in endless loop of attempts to start the container. Since this is single point of failure and job runs only once I would like to propose removal of these limits completely. Second option would be to use the `{{- toYaml .Values.bootstrap.ddldml.resources | nindent 10 }}` for final container as well.

### Event log with OOM error
(combined from similar events): Failed to create pod sandbox: rpc error: code = Unknown desc = container create failed: time="2020-10-21T16:25:24Z" level=warning msg="signal: killed" time="2020-10-21T16:25:24Z" level=error msg="container_linux.go:349: starting container process caused \"process_linux.go:365: sending config to init process caused \\\"write init-p: broken pipe\\\"\"" container_linux.go:349: starting container process caused "process_linux.go:365: sending config to init process caused \"write init-p: broken pipe\""

### DIFF
--- a/charts/influxdb-enterprise/templates/bootstrap-job.yaml
+++ b/charts/influxdb-enterprise/templates/bootstrap-job.yaml
@@ -61,10 +61,10 @@ spec:
         resources:
           requests:
             cpu: 100m
-            memory: 20Mi
+            memory: 50Mi
           limits:
             cpu: 100m
-            memory: 20Mi
+            memory: 50Mi
       {{ end }}
       {{- if .Values.bootstrap.ddldml.configMap }}
       - name: ddl
@@ -151,8 +151,8 @@ spec:
         resources:
           requests:
             cpu: 100m
-            memory: 20Mi
+            memory: 50Mi
           limits:
             cpu: 100m
-            memory: 20Mi
+            memory: 50Mi
 {{ end }}


rawkode:
Hi @pdrastil,

We cannot remove these resource limits, as it's very common for enterprise Kubernetes clusters to have validating webhooks to deny any workload without them in-fact, InfluxData has customers that require this and that's why they're enforced.

The hardcoded tasks have been tested to work, and the ddl/dml jobs have configurable limits as we cannot predict how much / how large your imports will be.

A PR would be accepted that raises the hard coded limits, nominally, if you're experiencing problems.

Let me know your thoughts

pdrastil:
Hi @rawkode, if this feature is required for the webhooks then I would propose the other option when we bump the hardcoded limit for final container or use same configuration section.

rawkode:
@pdrastil No problem. Do you want to bump them up, or would you like me to do so?

Can you test what limits work for your environment, please?

If you want to work together on this, DM me on the Community Slack and we'll sort it out.

Thanks

pdrastil:
@rawkode I would like to work on this with you via Slack because I have few more proposals I've discovered during the testing the deployment. Can you post here the invite? I'm currently getting 404 when clicking on the link in the repository.

rawkode:
CC @mhall119 - I ran into this NoSuchKey myself too, though it seems intermittent.

pdrastil:
Based on our Slack discussion limit was just bumped to 50Mi

rawkode (approved):

mhall119:
@pdrastil we're fixing the redirect link now, in the mean time you can join with https://join.slack.com/t/influxcommunity/shared_invite/zt-f8i7xcd2-zTRaIuPmE3yWYOFNOdJFLQ

sumo-drosiek (approved):

timhallinflux (approved):

PR #224, opened 2020-10-22
prefix secrets and configmaps with influxdb

pdrastil:
- Unified defaults for configuration objects so they can be clearly identified when namespace contains more deployments.

- Fixed naming mismatch for sample TLS secret and default in values.yaml

### DIFF
--- a/charts/influxdb-enterprise/example-resources.yaml
+++ b/charts/influxdb-enterprise/example-resources.yaml
@@ -1,14 +1,14 @@
 # apiVersion: v1
 # kind: Secret
 # metadata:
-#   name: license
+#   name: influxdb-license
 # stringData:
 #   json: YOUR JSON KEY
 ---
 apiVersion: v1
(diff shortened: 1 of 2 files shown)


rawkode:
Thanks for the PR, @pdrastil.

This is a welcomed improvement. I'm curious if we should use the Helm release name, which would also allow us to support multiple InfluxDB Enterprise deployments on a single cluster. Though perhaps that's not likely to happen?

Just a thought, I'm happy to approve these either way.

rawkode (approved):

pdrastil:
We could support Helm release but I wouldn't do it. Typically you would keep your licences and certificates in another system and create them only once for certain deployment. Also license is for given number of replicas so this is more for convenience when you are testing integration with other solutions and you just want sensible defaults.

sumo-drosiek (approved):
LGTM

PR #225, opened 2020-10-22
use pod anti-affinity to support HA deployment by default

pdrastil:
Enterprise deployments should run in HA setup. This PR adds sensible default configuration that try to deploy data and meta containers on different workers in K8s cluster if possible by default.

### DIFF
--- a/charts/influxdb-enterprise/values.yaml
+++ b/charts/influxdb-enterprise/values.yaml
@@ -59,7 +59,18 @@ meta:
     # repository: influxdb
   # nodeSelector: {}
   # tolerations: []
-  # affinity: {}
+  affinity:
+    podAntiAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
+      - weight: 1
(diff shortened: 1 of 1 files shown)


rawkode (changes requested):
Rather than baking this into the `values.yaml`, can we add it directly to the `Deployment` object?

IF a value is provided in the `values.yaml`, we use that instead.

pdrastil:
The reason why it's placed into values instead of hardcoded in StatefulSet is because you should allow modification of this configuration. I can imagine that someone would like to change this to `requiredDuringSchedulingIgnoredDuringExecution` or add additional rules with weights where to schedule pods in their cluster. This anti-affinity is just a starting point that should be good enough for most of the deployments.

rawkode (approved):

rawkode:
It's not a big deal, I can tidy up the logic afterwards if needed.

Thanks

sumo-drosiek (approved):
LGTM

Issue #227, opened 2020-10-25
[influxdb2] Install fails on influx setup


mproffitt:
I'm currently using the influxdb2 chart with Terraform and a AWS persistent volume

When upgrading/re-installing the chart, whilst maintaining the existing installation volume data, I am seeing the install fail with:

```
Error: failed post-install: job failed: DeadlineExceeded

  on ../common/helm/influxdb.tf line 1, in resource "helm_release" "influxdb":
   1: resource helm_release influxdb {
```

The failing step comes down to the execution of job-setup-admin `influx setup` task which returns 1 if the instance has already been configured. This causes the container used by the `job.setup-admin` to fail even though there isn't really an error.

```
root@test-host-influxdb2-0:/# influx setup -f --host http://test-host-influxdb2:8086 -o influxdata -b default -u admin -p ${INFLUXDB_PASSWORD} -t ${INFLUXDB_TOKEN}
Error: instance at "http://test-host-influxdb2:8086" has already been setup
See 'influx setup -h' for help
root@test-host-influxdb2-0:/# echo $?
1
```
Temporarily adding `|| echo` to the end of the influx setup command allows the chart to install correctly, however this is neither desirable, or sustainable

A more suitable solution would be to detect if influx has already been configured, in which case allow the job container to succeed, or alternatively to allow the job to be skipped entirely through a values setting?

rawkode:
I'll add an option to disable the job, would this resolve your issue @mproffitt ?

mproffitt:
It would indeed.

I think the method I'd submitted was an auto-skip but any option to allow the job to be bypassed if a re-install is required would be suitable and I'm happy for the onus to be back on the user to understand why the job should or should not run.

timhallinflux:
closed via #294

PR #228, opened 2020-10-25
#227 [influxdb2] Install fails on influx setup

mproffitt:
As described in #227, the influxdb2 helm chart will fail if influxdb2 has already been configured on the persistent volume store.

Although not fool-proof, the simplest way to overcome this failure is to execute `influx org list -n {{ .Values.adminUser.organisation }}` and only if that fails, move on to execute influx setup.

I believe this would work in all instances except where a user has deleted the configured `adminUser.organisation`

- [ ] CHANGELOG.md updated
- [x] Rebased/mergable
- [ ] Tests pass

### DIFF
--- a/charts/influxdb2/Chart.yaml
+++ b/charts/influxdb2/Chart.yaml
@@ -5,7 +5,7 @@ name: influxdb2
 description: A Helm chart for InfluxDB v2
 home: https://www.influxdata.com/products/influxdb-overview/influxdb-2-0/
 type: application
-version: 1.0.11
+version: 1.0.12
 maintainers:
   - name: rawkode
     email: rawkode@influxdata.com
--- a/charts/influxdb2/templates/job-setup-admin.yaml
+++ b/charts/influxdb2/templates/job-setup-admin.yaml
@@ -33,7 +33,10 @@ spec:
         args:
           - -c
           - |
-            influx setup -f \
+            influx org list \
+            -n {{ .Values.adminUser.organization }} \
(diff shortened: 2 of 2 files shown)


rawkode:
Thanks for taking a go at this.

I'm a little confused, as we use the install hook that should only be run for fresh installations.

`"helm.sh/hook": post-install`

This job shouldn't be created for upgrades.

mproffitt:
The scenario I found this in was when changing namespaces for the pod but I think it would trigger in any scenario whereby the pod was destroyed / replaced without the underlying filesystem being stripped back.

Another scenario I'm currently facing which would trigger this failure is migrating deployments between distinct clusters.

Perhaps the term "upgrades" is a little too ambiguous - I changed namespaces and changed versions simultaneously and then spent most of last night really confused as to why it stopped installing.

iamolegga:
Just faced the same issue, any updates?

ondrej-ivanko:
The script can work on most occasions, but not in following scenario. 

Suppose you have persistent volume claim enabled and it's created after you install influx2 release in your namespace. Influx is set up ok with script. 
Than you uninstall influx and install it again. If you made any changes to values.yaml adminUser on your second install of release, than all those changes are ignored. Because the influx is already set up in your persistent volume claim. You have to removed the pvc with kubectl and than can install again.

But that's a Kubernetes thing, I dont think this chart can do anything about it.

timhallinflux:
closing this in favor of #294

PR #232, opened 2020-11-06
Set Max Series

rohbharadwaj:


timhallinflux (approved):

charts/influxdb-enterprise/values.yaml
```diff
@@ -227,6 +227,7 @@ data:
     # annotations:
     # accessMode: ReadWriteOnce
     # size: 8Gi
+  maxseries: 1500000
```
michael-lee-mulesoft, line 230: What maxseries do we have in our cloud version? I don't think we have such a limit.

rawkode (changes requested):
Default value should be 1000000

This is the default when omitted

charts/influxdb-enterprise/values.yaml
timhallinflux: 1000000

rawkode:
No update since Nov 10th, 2020. Ping me if you'd like to get this going again.

Issue #240, opened 2020-11-25
[influxdb] Backup with persistence enabled fills up PVC


manu11th:
When using a PVC with `.Values.backup.persistence.enabled` it fills up after some time (depending on backup and PVC size) because for each backup a unique named folder is created with the timestamp `-portable /backup/"$(date +%Y%m%d%H%M%S)"`.

It would be beneficial to either remove the backup from the PVC after a successful upload or always use the same folder on the PVC to overwrite the last backup.
Another solution would be to use the `backupRetention` there (which is currently only visible in the values.yaml and not in the chart configuration documentation) and add a configuration point for the retention on the PVC when `.Values.backup.persistence.enabled`.

PR #241, opened 2020-11-25
[influxdb] Remove folders from PVC after uploading

manu11th:
Fixes influxdata/helm-charts#240
When a PVC is used before the data upload then it fills up quickly after some backups because always a new timestamped folder is created.
This PR fixes the problem by removing all folders located under `/backup/*`after uploading them.

### DIFF
--- a/charts/influxdb/Chart.yaml
+++ b/charts/influxdb/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: influxdb
-version: 4.8.9
+version: 4.8.10
 appVersion: 1.8.0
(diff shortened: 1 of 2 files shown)


rawkode:
Hey @manu11th,

Thank you for the PR. Could you please add `-e` to the `sh` arguments? This should stop the backups being removed if the sync to an external location fails.

Thanks

manu11th:
Hey @rawkode,

Thanks for the hint, I have added the flag.

timhallinflux (approved):

rawkode (approved):

PR #244, opened 2020-12-03
Adds persistence.existingClaim value to Chronograf chart

patrickjmcd:
In order to accommodate PersistentVolumeClaims that have already been created, I've added in the `persistence.existingClaim` value in [the same way that Grafana does it](https://github.com/grafana/helm-charts/blob/74624247a5676bd8d2c5460c47bd984aaad35712/charts/grafana/templates/_pod.tpl#L417).

This should allow users to use other storage methods (NFS in my case) to store Chrongraf configuration data.

I have tested this locally via `helm install...`

### DIFF
--- a/charts/chronograf/Chart.yaml
+++ b/charts/chronograf/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: chronograf
-version: 1.1.22
+version: 1.1.23
 appVersion: 1.8.9.1
 description: Open-source web application written in Go and React.js that provides
   the tools to visualize your monitoring data and easily create alerting and automation
--- a/charts/chronograf/Chart.yaml
+++ b/charts/chronograf/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: chronograf
-version: 1.1.19
+version: 1.1.20
 appVersion: 1.8.8
 description: Open-source web application written in Go and React.js that provides
   the tools to visualize your monitoring data and easily create alerting and automation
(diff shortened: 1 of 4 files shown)


rawkode (approved):

rawkode (dismissed):

charts/chronograf/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: chronograf
-version: 1.1.19
+version: 1.1.20
```
timhallinflux, line 3: just need to bump this forward.  We've had another merge since...

charts/chronograf/Chart.yaml
```diff
@@ -1,6 +1,6 @@
...
 appVersion: 1.8.8
```
timhallinflux, line 4: 1.8.9 is now out too.  Let's bump that forward as well.

charts/chronograf/Chart.yaml
patrickjmcd: bumped in 9310ec2

charts/chronograf/Chart.yaml
patrickjmcd: bumped in 9310ec2

timhallinflux (dismissed):

rawkode:
@patrickjmcd I think your recent commit has injected some weird formatting changes?

rawkode:
No updates for a while. Please feel free to ping me if you'd like to work on this and get it merged.

PR #246, opened 2020-12-14
[telegraf] Fix: Cannot define inputs without health check

philomory:
In the current implementation, a misplaced `{{- end }}` in the chart causes the `{{ template "inputs" .Values.config.inputs -}}` call to be inside the `{{- if .Values.metrics.health.enabled }}` conditional block.

This fixes that.

### DIFF
--- a/charts/telegraf/Chart.yaml
+++ b/charts/telegraf/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: telegraf
-version: 1.7.33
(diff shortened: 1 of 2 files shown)


rawkode (dismissed):

philomory:
@rawkode who else needs to approve this before it can get merged?

timhallinflux (dismissed):

timhallinflux (approved):

PR #247, opened 2020-12-16
Added missing securityContext options

abhi1693:


rawkode:
@abhi1693 You've kind of "slipped" the mountPath changes into this, but this will cause InfluxDB not to persist.

The following flag needs added to the `influxd` process: `influxd --engine-path=PATH`

https://docs.influxdata.com/influxdb/v2.0/reference/config-options/

abhi1693:
@rawkode so, should I add the command to the statefulset? Also, does it need to have any other parameters if I were to override CMD of the docker?

abhi1693:
@rawkode All fixed as per your suggestion

rawkode:
Is there duplication with #252 ?

abhi1693:
I'll close this one in favor of #252

PR #250, opened 2020-12-16
Reorder conditional end to allow inputs to render

### LINKED ISSUE #249: telegraf chart: inputs no longer rendered
It appears that #237 mistakenly included the inputs inside of a new conditional, resulting in no inputs being rendered after upgrading to chart 1.7.33.

The offending lines are here: https://github.com/influxdata/helm-charts/blob/2ee220c6557c8538dc465ebb8560fde200c47d8a/charts/telegraf/templates/configmap.yaml#L22-L23

The inputs should be rendered unconditionally, so the order of these lines just needs to be switched so that the inputs come outside the conditional's  `{{- end}}`.

timhallinflux:
fixes #249

### DIFF
--- a/charts/telegraf/Chart.yaml
+++ b/charts/telegraf/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: telegraf
-version: 1.7.33
+version: 1.7.34
 appVersion: 1.16
 deprecated: false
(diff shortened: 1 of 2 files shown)


rawkode (approved):

rawkode:
NIce catch, @timhallinflux

PR #253, opened 2021-01-07
Supply TOML in inputs_override_config

stblassitude:
The translation of Helm YAML to Telegram TOML in the chart has
limitations. Allow arbitrary TOML for the input section by putting a
multi-line string under config.inputs_override_config.

Example:
```
config:
  inputs_override_config:

      [[inputs.snmp]]
        agents = [ "udp://192.168.0.2:161" ]
        community = "public"
        version = 1
        name_override = "my-ups"
        interval = "60s"
      [[inputs.snmp.field]]
        name = "model"
        oid = "1.3.6.1.4.1.318.1.1.1.1.1.1.0"
        is_tag = true
      [[inputs.snmp.field]]
        name = "input_voltage"
        oid = "1.3.6.1.4.1.318.1.1.1.3.2.1.0"
      [[inputs.snmp.field]]
        name = "input_frequency"
        oid = "1.3.6.1.4.1.318.1.1.1.3.2.4.0"

```


stblassitude:
I don't understand the linting error. Am I supposed to bump the chart version number?

mhall119:
@timhallinflux is the `lint-test` blocked on something?

charts/telegraf/templates/configmap.yaml
```diff
@@ -20,6 +20,10 @@ data:
       [[outputs.health.contains]]
         field = "buffer_size"
     {{- end }}
+    {{- if .Values.config.inputs_override_config }}
```
timhallinflux, line 23: So a question about this approach.

Is there a reason why this is input specific?  This really could be any part of the Telegraf config, no?  global, input, output, processor, aggregator??

Perhaps we should alter this to be additional_config and have it appended at the end of the configmap?

stblassitude: This change was the minimal change to support my use case. If you'd rather have a different implementation, let me know and I will try to update it. In my view, the attempt to convert a dict to the TOML cannot work correctly, so I would even suggest to deprecate it and simply replace it with a (multi-line) string of the TOML. The code in _helpers.tpl for define "inputs" is way too complicated and brittle, I think. I have not looked at the other sections, but I suspect the case can be made there as well.

PR #254, opened 2021-01-12
Add loadBalancerSourceRanges support, refactor config map

szymonpk:
Add support for service.spec.loadBalancerSourceRanges. We can control
security group entries for AWS load balancers. Chart supported this in
previous versions (3.x).

Refactor how Helm handles long numbers. Due to a bug, during YAML to
JSON conversion, Helm may write long numbers in scientific notation.
InfluxDB does not support that. For convenience, move config map section
generation to the Helm helper.


szymonpk:
There is no CHANGELOG.md for this chart, so nothing to change/update.

charts/influxdb/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: influxdb
-version: 4.8.10
+version: 4.9.10
```
szymonpk, line 3: I am not sure what version I should use, according semver it should be 4.9.0.

rawkode:
Happy to accept the load balancer changes, but not the config map break.

Please feel free to clean this up and tag me

rawkode (changes requested):
CIDR range is OK. Please don't modify the configmap

naseemkullah:
> Happy to accept the load balancer changes, but not the config map break.

Looks like a nice factoring out with one logical changes which is to `| int64` numeric values to avoid having them converted to scientific notation apparently. Is that breaking?

rawkode:
> > Happy to accept the load balancer changes, but not the config map break.
> 
> Looks like a nice factoring out with one logical changes which is to `| int64` numeric values to avoid having them converted to scientific notation apparently. Is that breaking?

Yes, as it assumes that values can only be strong, bool, or int. There’s no clause for arrays or floats. 

TOML wrangling is a mess in Helm and I’d rather discourage using yaml to generate it. Of course, if you wish to add some Rego to confirm it works with a varied and complete configuration, then by all means do. 

I’ve added a basic test case to build into CI in an unrelated PR

https://github.com/influxdata/helm-charts/pull/281/files

charts/influxdb/templates/_helpers.tpl
```diff
@@ -61,3 +61,22 @@ Create the name of the service account to use
...
+{{- $tp := typeOf $value }}
+{{- if eq $tp "string" }}
+{{ $key }} = {{ $value | quote }}
+{{- else if eq $tp "bool" }}
+{{ $key }} = {{ $value }}
+{{- else }}
+{{ $key }} = {{ $value | int64 }}
+{{- end  }}
+{{- end }}
+{{- end -}}
```
naseemkullah, line 82:
```suggestion
{{- if eq $tp "string" }}
{{ $key }} = {{ $value | quote }}
{{- else }}
{{ $key }} = {{ $value }}
{{- end }}
{{- end }}
```

Factoring out is great! But let's address this logical change in a separate PR as there are concerns over the arrays, floats as mentioned in https://github.com/influxdata/helm-charts/pull/254#issuecomment-787414888

OOC, is there a value in your current influxdb config that is turning into scientific notation today @szymonpk ?

charts/influxdb/templates/_helpers.tpl
```diff
@@ -61,3 +61,22 @@ Create the name of the service account to use
     {{ default "default" .Values.serviceAccount.name }}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Format service config key values
+Numeric values require special handling due to the Helm upstream issue, otherwise they may be converted to scientific
```
naseemkullah, line 67:
```suggestion
todo: Numeric values may require special handling due to the Helm upstream issue, otherwise they may be converted to scientific
```

charts/influxdb/templates/configmap.yaml
```diff
@@ -105,8 +49,10 @@ data:
       {{- $tp := typeOf $value }}
       {{- if eq $tp "string" }}
       {{ $key }} = {{ $value | quote }}
-      {{- else }}
+      {{- else if eq $tp "bool" }}
       {{ $key }} = {{ $value }}
+      {{- else }}
+      {{ $key }} = {{ $value | int64 }}
```
naseemkullah, line 55:
```suggestion
      {{- else }}
      {{ $key }} = {{ $value }}
```

charts/influxdb/templates/_helpers.tpl
szymonpk:
```
    data:
        query-log-enabled: false
        cache-max-memory-size: 1073741824
```

IIRC, influxv1 does not accept quoted value here.

szymonpk:
> > > Happy to accept the load balancer changes, but not the config map break.
> > 
> > 
> > Looks like a nice factoring out with one logical changes which is to `| int64` numeric values to avoid having them converted to scientific notation apparently. Is that breaking?
> 
> Yes, as it assumes that values can only be strong, bool, or int. There’s no clause for arrays or floats.
> 
> TOML wrangling is a mess in Helm and I’d rather discourage using yaml to generate it. Of course, if you wish to add some Rego to confirm it works with a varied and complete configuration, then by all means do.
> 
> I’ve added a basic test case to build into CI in an unrelated PR
> 
> https://github.com/influxdata/helm-charts/pull/281/files

@rawkode I have found a case for arrays `access-log-status-filters = []` ([doc](https://docs.influxdata.com/influxdb/v1.8/administration/config/)), but I do not see any values that can be floats. Do I miss something?

szymonpk:
@rawkode @naseemkullah I have dropped 'int64' shenanigans, but left the helper, I can drop it too if you want. I have also rebased the code with master.

naseemkullah (approved):
Thanks @szymonpk

PR #270, opened 2021-02-15
Make readiness and liveness checks configurable

fredericOfTestfabrik:
The choosen defaults for readiness/liveness checks may trigger unnecessary container restarts, because timeouts may be too low. 

I simply introduced a readiness/liveness config options and used the in the statefulset.yaml. 

I tested my changes by locally installing the chart and checking if the values were set correctly.

- [ ] CHANGELOG.md updated
- [ ] Rebased/mergable
- [ ] Tests pass

### DIFF
--- a/charts/influxdb2/Chart.yaml
+++ b/charts/influxdb2/Chart.yaml
@@ -5,7 +5,7 @@ name: influxdb2
 description: A Helm chart for InfluxDB v2
 home: https://www.influxdata.com/products/influxdb-overview/influxdb-2-0/
 type: application
-version: 1.0.16
+version: 1.1.1
 maintainers:
   - name: rawkode
     email: rawkode@influxdata.com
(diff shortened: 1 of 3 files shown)


rawkode:
Hi @fredericOfTestfabrik,

Thanks. This looks great. I'm OK with the timeouts being configurable, but don't believe we should make the port or path configurable.

Can you remove those options, please?

fredericOfTestfabrik:
I'm a fan of make everything available to the end user but I guess you are right, in this case it doesn't really make sense.

I removed the options as requested.

rawkode (dismissed):

timhallinflux (changes requested):
Chart version needs a bump

fredericOfTestfabrik:
Sorry. Unfortunately I was very busy. I bumped the version as requested.

Tonow:
I think this PR is more complete https://github.com/influxdata/helm-charts/pull/381

timhallinflux:
closing in favor of #381

PR #274, opened 2021-02-19
Allow passing extra flags/arguments to influxd

asherf:
This will allow users of this chart to configure/set any of the command line flags defined in https://docs.influxdata.com/influxdb/v2.0/reference/cli/influxd/run/#flags

### DIFF
--- a/charts/influxdb2/Chart.yaml
+++ b/charts/influxdb2/Chart.yaml
@@ -5,7 +5,7 @@ name: influxdb2
 description: A Helm chart for InfluxDB v2
 home: https://www.influxdata.com/products/influxdb-overview/influxdb-2-0/
 type: application
-version: 1.0.16
+version: 1.0.17
 maintainers:
   - name: rawkode
     email: rawkode@influxdata.com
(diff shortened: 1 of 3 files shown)


rawkode:
Everything in InfluxDB can be configured through configuration file and environment variables, I'm not sure if adding new flags to the command line is a great way forward.

Do you have a use-case that can only be handled this way?

asherf:
Ah. Got it.
Sorry, I am new to influxdb... 
I'll use the config file.
Thanks!

asherf:
@rawkode 
Actually, after looking at the docs, influxd expects the config file to be in the same directory as the binary, which is tricky to do in a k8s/container situation (since config maps need to be mounted on their own paths)
alternatively, environment variables or can be used to config or to specify a different location for the config file.
I am happy to do any of those solutions.
Any feedback on which of those will be preferred  ?

rawkode:
Hi @asherf You can set an environment variable to specify where the config files are located.

`INFLUXD_CONFIG_PATH: /path/to/custom/config/directory`

asherf:
FWIW, the helm chart doesn't currently allow specifying additional environment variables.
so in order to support that using a config file in a different location, that needs to be supported (or at least added).
After looking at the docs, adding command line arguments still seemed to be the easiest way to add config options.
Especially since InfluxDB in a way where it doesn't care where the config options are coming from and this chart already a command line argument to configure the engine path.

rawkode:
You are right!

I'll make this possible #282

PR #275, opened 2021-02-20
[influxdb] fix ingress className set to string nil

FlipEnergy:
Due to a recent change, a default installation will cause ingress className to be set to the string 'nil'. This PR fixes that to be empty string.

```
kind: Ingress
apiVersion: extensions/v1beta1
...
spec:
  ingressClassName: nil
```

Signed-off-by: Dennis Zhang <dennis.zhang.nrg@gmail.com>

- [ ] CHANGELOG.md updated
- [ ] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/charts/influxdb/Chart.yaml
+++ b/charts/influxdb/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: influxdb
-version: 4.9.13
+version: 4.9.14
 appVersion: 1.8.4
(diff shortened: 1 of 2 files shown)


rawkode:
Thank you for submitting this @FlipEnergy. Unfortunately, I've opted for #278, because `null` is the correct YAML/Helm.

Thanks though!

PR #279, opened 2021-02-23
[influxdb2] Add possibility to use existing secret for admin password

florianrusch:
I've missed the possibility to use existing secret for the admin password and the token.
For the implementation I followed the Helm Chart for InfluxDB 1.


rawkode (approved):
Great PR and welcomed addition. Thank you.

timhallinflux (approved):

PR #281, opened 2021-02-26
Feat/allow infuxdb2 tag override

rawkode:


rawkode:
Hopefully this gets everything passing. There's a discrepency between the Quay image and the Hub image, annoyingly.

rawkode:
Decided to do a major bump, as there's some risk switching the images.

timhallinflux (approved):

timhallinflux:
Lots of changes were made to the final image that was posted to DockerHub. This was in order to facilitate upgrades from 1.x to 2.x. 

Worth reviewing all the gory details here: https://hub.docker.com/_/influxdb

charts/influxdb2/Chart.yaml
```diff
@@ -5,7 +5,7 @@ name: influxdb2
 description: A Helm chart for InfluxDB v2
 home: https://www.influxdata.com/products/influxdb-overview/influxdb-2-0/
 type: application
-version: 1.0.17
+version: 2.0.0
```
danxmoran, line 8: Since there's no `2.0.0` on DockerHub, should this be `2.0` or `2.0.4`? That way the fallback case will work.

PR #285, opened 2021-03-10
[telegraf] Add envFromSecret to pass sensitive data into environment

ymatsiuk:
Add ability to pass sensitive environment variables to telegraf environment.
Specific use case might be passing `$INFLUX_TOKEN` or any other API token:

Example `values.yaml` will look like this:
```
...
envFromSecret: "telegraf-tokens"
env:
  - name: HOSTNAME
    value: "telegraf-polling-service"
  - name: INFLUX_ORG
    value: influxdata
  - name: INFLUX_HOST
    value: https://example.com
...
config:
  agent:
...
    hostname: "$HOSTNAME"
    omit_hostname: false
  outputs:
    - influxdb_v2:
        urls:
          - "${INFLUX_HOST}"
        token: "${INFLUX_TOKEN}"
        organization = "${INFLUX_ORG}"
...
```

### DIFF
--- a/charts/telegraf/templates/deployment.yaml
+++ b/charts/telegraf/templates/deployment.yaml
@@ -41,6 +41,11 @@ spec:
         {{- end }}
         env:
 {{ toYaml .Values.env | indent 8 }}
+{{- if .Values.envFromSecret }}
+        envFrom:
+          - secretRef:
+              name: {{ .Values.envFromSecret . }}
+{{- end }}
         volumeMounts:
         - name: config
           mountPath: /etc/telegraf
--- a/charts/telegraf/Chart.yaml
+++ b/charts/telegraf/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: telegraf
 version: 1.8.0
-appVersion: 1.18.0
+appVersion: 1.18.3
 deprecated: false
 description: Telegraf is an agent written in Go for collecting, processing, aggregating, and writing metrics.
 keywords:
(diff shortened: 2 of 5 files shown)


ymatsiuk:
@rawkode @timhallinflux Could you take a look, please 🙏🏻 . Thanks in advance.

charts/telegraf/templates/deployment.yaml
```diff
@@ -41,6 +41,11 @@ spec:
         {{- end }}
         env:
 {{ toYaml .Values.env | indent 8 }}
+{{- if .Values.envFromSecret }}
+        envFrom:
+          - secretRef:
+              name: {{ tpl .Values.envFromSecret . }}
```
rawkode, line 47: This should be `.Values.envFromSecret` rather than using `tpl`, as the value in the values file won't be a template itself.

rawkode (changes requested):
Commented

charts/telegraf/templates/deployment.yaml
ymatsiuk: @rawkode you're totally correct, I somewhat overthought this 🙂 Now I see we don't have any secret references in `values.yaml`

ymatsiuk:
I pushed the fix for the errors that lint threw, to make the checks pass (these are in separate commit though).

timhallinflux (approved):

dcarrier:
Thank you for making these changes this will be very useful. Any chance we can get movement on merging this?

PR #291, opened 2021-03-23
Updated app version to v1.18.0

abhi1693:


nsteinmetz:
It's not enough to upgrade telegraf version - you also have to update image version in values.yaml files.

https://github.com/influxdata/helm-charts/blob/master/charts/telegraf/values.yaml#L9

PR #290, opened 2021-03-23
Add container environment variable configuration

rmeharg:
Allow configuration of container environment variables.

For greater flexibility and configuration of options found that are not explicitly exposed in the helm chart 

https://docs.influxdata.com/enterprise_influxdb/v1.8/administration/config-data-nodes/#data-node-configuration-settings

Our use case is for INFLUXDB_HTTP_FLUX_ENABLED

Issue #288, opened 2021-03-18
influxdb2 helm chart need update to dockerhub version of influxdb2


kOstovic:
Current Helm chart reference quay.io version of influxdb2 docker. Helm need's to be updated to official release from dockerhub. Moreover this isn't just update image and chart names in templates, but also update of job section and liveness/readiness probe because it fails on them too. Also there is no way to use env variables which means some things like upgrade is not supported. 

Therefor there is need to refactor job section and add env section in deployment of stateful set

timhallinflux:
closed via #294

PR #283, opened 2021-03-01
Always deploy job-setup-admin pod on a linux node

Mierpo:
Hybrid clusters with windows and linux nodes would potentially
deploy the pod on a windows worker. Using a nodeSelector for the
admin setup job fixes this.

- [ ] CHANGELOG.md updated
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

related to #256

Issue #282, opened 2021-02-26
Allow Environment Variables / InfluxDB Data Dir Configuration for InfluxDB2


rawkode:

timhallinflux:
closed via #294

PR #294, opened 2021-04-03
influxdb2 - persist bolt-path in volume. use docker hub.

### LINKED ISSUE #293: [influxdb2] BoltDB not persisted after restart with custom mountPath
# Summary

I recently upgraded to influxdb2 using this chart. As I was doing a backup restore from a migrated influxdb 1.8 installation I set the `mountPath` to `/var/lib/influxdb2` to match the docker image.

A week or so later once the pod restarted due to a node scale down, my influxdb was "empty" and I was met with the onboarding splash screen.

After some digging I can see that my data is at the correct location but the bolt db is in `/root/.influxdb2`.

I think my confusion came from the helm chart using quay.io AND v1.8. It isn't currently possible to use the docker hub version as this chart forces a `v` in the docker tag. Previous me had already seen this as an issue, I just didn't realise that the bolt-path was a thing 🤦 

## Configuration

You can see my configuration here:

https://github.com/Smirl/nest_temperature_forwarder/blob/master/deploy/nest-temperature-forwarder/values.yaml#L102-L121

## Possible Fix

Set the `--bolt-path` to be inside the persistent volume `mouthPath`

## Data recovery

Does anyone know how I can "recreate" a boltdb so I don't have data loss. I have the `/data`, `/wal`, etc. from the `--engine-path` but not a proper backup? Any help on this is appreciated.

Smirl:
# Summary

closes #293 

- use docker hub rather than quay.io
- ensure that boltdb is also persisted in the Volume if mouthPath is set

### DIFF
--- a/charts/influxdb2/templates/statefulset.yaml
+++ b/charts/influxdb2/templates/statefulset.yaml
@@ -32,12 +32,14 @@ spec:
         {{- end }}
       containers:
         - name: {{ .Chart.Name }}
-          image: "{{ .Values.image.repository }}:v{{ .Chart.AppVersion }}"
+          image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
             - influxd
             - --engine-path
             - {{ .Values.persistence.mountPath }}
+            - --bolt-path
+            - {{ .Values.persistence.mountPath }}/influxd.bolt
           ports:
             - name: {{ .Values.service.portName }}
               containerPort: 8086
--- a/charts/influxdb2/values.yaml
+++ b/charts/influxdb2/values.yaml
@@ -1,5 +1,6 @@
 image:
-  repository: quay.io/influxdb/influxdb
+  repository: influxdb
+  # tag:
   pullPolicy: IfNotPresent
 
 podAnnotations: {}
@@ -64,7 +65,7 @@ persistence:
   # storageClass: "-"
   accessMode: ReadWriteOnce
   size: 50Gi
-  mountPath: /root/.influxdbv2
+  mountPath: /var/lib/influxdbv2
 
 service:
   type: ClusterIP
--- a/charts/influxdb2/Chart.yaml
+++ b/charts/influxdb2/Chart.yaml
@@ -5,7 +5,7 @@ name: influxdb2
 description: A Helm chart for InfluxDB v2
 home: https://www.influxdata.com/products/influxdb-overview/influxdb-2-0/
 type: application
-version: 1.1.0
+version: 1.2.0
 maintainers:
   - name: rawkode
     email: rawkode@influxdata.com


Smirl:
Ok, so it was a bit more involved than I first thought. However, this should now cover the move to dockerhub and closes a bunch of issues in the process. I see that there are some old PRs that have tried to address these issues. Thanks to those contributors too.

Tag can now be specified. Falls back to AppVersion if not set:
closes: https://github.com/influxdata/helm-charts/issues/147
closes: https://github.com/influxdata/helm-charts/pull/281

Uses the docker hub entrypoint to set up influx. This is idempotent so won't have errors setting up the db twice:
closes: https://github.com/influxdata/helm-charts/issues/227
closes: https://github.com/influxdata/helm-charts/pull/228

Extra environment variables can now be passed to configure InfluxDB:
closes: https://github.com/influxdata/helm-charts/issues/238
closes: https://github.com/influxdata/helm-charts/issues/282
closes: https://github.com/influxdata/helm-charts/pull/290

Setup job is removed in favour of docker entrypoint.sh:
closes: https://github.com/influxdata/helm-charts/issues/256
closes: https://github.com/influxdata/helm-charts/pull/283

Docker hub image is used by default:
closes: https://github.com/influxdata/helm-charts/issues/288

We use the `lookup` function to reuse the password on upgrade.
Required upgrade of helm tools in github-actions.
closes: https://github.com/influxdata/helm-charts/pull/189

charts/influxdb2/templates/statefulset.yaml
```diff
@@ -32,16 +32,62 @@ spec:
         {{- end }}
       containers:
         - name: {{ .Chart.Name }}
-          image: "{{ .Values.image.repository }}:v{{ .Chart.AppVersion }}"
+          image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
```
danmx, line 35: I'd consider dropping it since it's not really following standard practices. Using `non-values` to configure default behaviour of a chart template is counter intuitive

charts/influxdb2/values.yaml
```diff
@@ -1,5 +1,6 @@
 image:
-  repository: quay.io/influxdb/influxdb
+  repository: influxdb
+  # tag:
```
danmx, line 3: Did you consider using `2.0.4-alpine` as a default?

charts/influxdb2/templates/statefulset.yaml
```diff
@@ -32,16 +32,62 @@ spec:
...
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          args:
```
danmx, line 37: extra `args` should be configurable, for example to configure [a log level in `entrypoint.sh`](https://github.com/influxdata/influxdata-docker/blob/368e523/influxdb/2.0/entrypoint.sh#L28-L35)

Smirl: `influxd print-config --key-name log-level` should take this from `INFLUXD_LOG_LEVEL`

charts/influxdb2/values.yaml
Smirl: I will leave this up to the maintainers.

Smirl:
@timhallinflux Would you mind taking a look at this one? It should hopefully fix up a bunch of open issues and close some other open PRs at the same time. Thanks!

danxmoran (approved):
New config setup LGTM

charts/influxdb2/values.yaml
timhallinflux: yeah...there are two options here `2.0-alpine` which means the maintenance revisions will be automatically pulled.
or `2.0.4-alpine` which means for every maintenance release, we'll need to update the chart.

But, we need this to be tagged...latest is undesirable.

timhallinflux:
@Smirl -- need the CLA signed. https://www.influxdata.com/legal/cla/

charts/influxdb2/values.yaml
Smirl: As it is it will default to the charts app version. I will change it to only be configured with the tag value and use 2.0.4-alpine

charts/influxdb2/templates/statefulset.yaml
```diff
@@ -32,16 +32,62 @@ spec:
...
+
+            # Path to the BoltDB database.
+            - name: INFLUXD_BOLT_PATH
+              value: {{ .Values.persistence.mountPath }}/influxd.bolt
+            # Path to persistent storage engine files where InfluxDB stores all Time-Structure Merge Tree (TSM) data on disk.
+            - name: INFLUXD_ENGINE_PATH
+              value: {{ .Values.persistence.mountPath }}
+
+            # Extra environment variables from .Values.env
+            {{- with .Values.env }}
```
rmeharg, line 87: Can you add this to the other charts as well, specifically influxdb-enterprise?

Smirl: Do you mind if we raise a separate PR for this as this one is already quite big?

rmeharg: Yes, it is already #290, just making sure it doesn't get closed because of this PR

charts/influxdb2/templates/statefulset.yaml
```diff
@@ -32,16 +32,62 @@ spec:
...
+                  {{ else }}
+                  name: {{ template "influxdb.fullname" . }}-auth
+                  {{- end }}
+                  key: admin-token
+
+            # Path to the BoltDB database.
+            - name: INFLUXD_BOLT_PATH
+              value: {{ .Values.persistence.mountPath }}/influxd.bolt
+            # Path to persistent storage engine files where InfluxDB stores all Time-Structure Merge Tree (TSM) data on disk.
+            - name: INFLUXD_ENGINE_PATH
```
danmx, line 83: that can be addressed in the `config.yml`:

```yaml
bolt-path: /var/lib/influxdb2/influxd.bolt
engine-path: /var/lib/influxdb2/engine
```

Smirl: There is no config file mounted and all options can be environment variables. I have seen in the past the maintainers preferred environment variables so I'll stick with that for consistency.

timhallinflux (approved):

danxmoran (approved):

rawkode (approved):

rawkode:
Great work, @Smirl

PR #298, opened 2021-04-12
Update backup-cronjob.yaml

maksagit:
Added nodeSelector for linux image: microsoft/azure-cli


rawkode:
What happens if the `nodeSelector` value doesn't exist,  @maksagit ?

PR #306, opened 2021-04-25
[telegraf] Create a script to be used by telegraf exec input plugin

anachristofaro:
Optionally create a script to be used by telegraf exec input plugin.

The main purpose is to allow the same script to be reused by more than one [[inputs.exec]] for different query parameters based on bash arguments. In addition, it also allows the user to enter a set of commands to be performed by the [[inputs.exec]] in a more fashion way.


rawkode (commented):
You should use the volume Mount options to inject a script. A helper in values won't be accepted

PR #342, opened 2021-07-08
Update template to use float if float is specified

imranismail:
This fixes an error with toml expecting float value for statsd plugin

```yaml
- statsd:
    service_address: :8125
    percentiles:
      - !!float 50
      - !!float 95
      - !!float 99
    metric_separator: _
    allowed_pending_messages: 10000
    percentile_limit: 1000
```


alespour:
Percentiles float were solved in #396

PR #347, opened 2021-07-28
[telegraf] Add support for sidecar and init containers

lawliet89:
Add ability to add sidecar and init containers for Telegraf. Tested on my own deployment.


alespour:
Fixes #395

PR #352, opened 2021-08-24
influxdb: support v1 ingress

jonkerj:
InfluxDB's Helm chart was lacking v1 ingress support, which makes it impossible to deploy on Kubernetes 1.22. This PR makes helm auto-detect if v1 ingress is possible.

### DIFF
--- a/charts/influxdb/Chart.yaml
+++ b/charts/influxdb/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: influxdb
-version: 4.10.0
+version: 4.10.1
 appVersion: 1.8.6
 description: Scalable datastore for metrics, events, and real-time analytics.
 keywords:
--- a/charts/influxdb/templates/ingress.yaml
+++ b/charts/influxdb/templates/ingress.yaml
@@ -1,5 +1,9 @@
 {{- if .Values.ingress.enabled -}}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
+apiVersion: networking.k8s.io/v1
+{{- else }}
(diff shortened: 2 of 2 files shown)


jonkerj:
The pipeline seems to fail on something else, anything I can do?

jonkerj:
@timhallinflux any chance you could take a look at this PR? Looks like you are an active dev here, and this issue is complicating influxdb deployment on recent k8s.

alespour:
Closes #353.

alespour (approved):
LGTM

ivankudibal (approved):

bentatham:
This might be silly, but we are still on k8s 1.18 which does not support `networking.k8s.io/v1`, and somehow the capabilities check is not working properly for me:

```
$ helm upgrade --namespace influx --version 4.10.1 --install --values values.yaml --set-string image.tag=1.8.10-alpine influxdb influxdata/influxdb
Error: UPGRADE FAILED: unable to recognize "": no matches for kind "Ingress" in version "networking.k8s.io/v1"
```
(using helm 3.7.0 with kubernetes v1.18.20-eks-8c579e)

Any ideas why the check `{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}` isn't working for me to generate the correct ingress apiVersion?

alespour:
@bentatham Apparently `{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}` check is not sufficient. Despite changes in k8s netwroking 1.18, it still does not support Ingress v1 (https://github.com/kubernetes/kubernetes/issues/90077).
The check should be more specific: `{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}`

Issue #119, opened 2020-06-30
[telegraf]: unable to pass `processor.converter.field` values


rorysedgwick:
**Telegraf Helm chart version: 1.7.7**
**Telegraf Docker image 1.13-alpine**
**Amazon Linux on EKS**

We want to convert a field (`system_location_sectionFloorNumber`) into a tag. The following config works when passed as direct `yaml` as part of a `telegraf.conf` file:

```
[[processors.enum]]
  order = 1
  [[processors.enum.mapping]]
    dest = "system_location_sectionFloorNumber"
    field = "system_location_section"
    [processors.enum.mapping.value_mappings]
        1st = 1
        2nd = 2
        3rd = 3
        4th = 4
        5th = 5
        Basement = -1
        Ground = 0
[[processors.converter]]
  order: 2
  [[processors.converter.fields]]
    tag: [ "system_location_sectionFloorNumber" ]
```

When included as the following, as part of the `values.yaml` that is passed to the Helm chart of the Telegraf Docker image, it returns the following error:
```
  processors:
    - enum:
        order: 1
        mapping:
          field: system_location_section
          dest: system_location_sectionFloorNumber
          value_mappings:
            2nd: 2
            1st: 1
            Ground: 0
            Basement: -1
    - converter:
        order: 2
        fields:
          tag: system_location_sectionFloorNumber
```

**`E! [telegraf] Error running agent: Error parsing /etc/telegraf/telegraf.conf, line 39: (converter.Converter.Fields) cannot unmarshal TOML array table into converter.Conversion (need slice)`**

This is the case whether the tag key is passed as
`tag: [ system_location_sectionFloorNumber ]`
```
tag:
  - system_location_sectionFloorNumber
```
etc

russorat:
Taking your telegraf TOML and converting it to YAML, i get the following (there were also some parse errors in your toml). Could you try this?

I used this: https://toolkit.site/format.html

```
processors:
  enum:
    - order: 1
      mapping:
        - dest: system_location_sectionFloorNumber
          field: system_location_section
          value_mappings:
            1st: 1
            2nd: 2
            3rd: 3
            4th: 4
            5th: 5
            Basement: -1
            Ground: 0
  converter:
    - order: 2
      fields:
        - tag:
            - system_location_sectionFloorNumber
```

lawliet89:
I am facing the same issue. Taking what @russorat has suggested, the chart renders:

```toml
    [[processors.0]]
      fields = [
      ]
[...]

PR #358, opened 2021-09-12
Allow for raw toml configs for telegraf

foobaar:
There are use cases where we the yaml->toml conversion does not work with the current template.
This allows for all the edge cases.


alespour:
Apologies for the delayed response. Could you please rebase, and bump chart version? Thank you.

alespour:
Closes #119 
Closes #429

alespour:
I've found that similar config override exists in telegraf-ds since #239, so I've just changed the key for the sake of certain consistency though snake case is used rarely, and added test file for CI.

Issue #310, opened 2021-04-30
[telegraf-ds] ServiceAccount bearer token points at wrong location


vzabawski:
According to [documentation](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/), service account token should be pointing at `/var/run/secrets/kubernetes.io/serviceaccount` instead of `/run/secrets/kubernetes.io/serviceaccount/token`. Current config produces the following error:
```2021-04-30T15:01:49Z E! [telegraf] Error running agent: could not initialize input inputs.kubernetes: open /run/secrets/kubernetes.io/serviceaccount/token: no such file or directory```

Replacing the value at configmap fixes pods.

alespour:
Solved by #361

PR #311, opened 2021-04-30
[telegraf-ds] Fix ServiceAccount token mounth path

vzabawski:
Fix issue when a pod with telegraf-ds can't start due to absence of bearer token.

Resolves issue #310

- [ ] CHANGELOG.md updated
- [x] Rebased/mergable
- [ ] Tests pass
- [x] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

PR #361, opened 2021-09-17
Fix ServiceAccount token mounth path

vzabawski:
Fix issue when a pod with telegraf-ds can't start due to absence of bearer token.

Resolves issue #310

- [ ] CHANGELOG.md updated
- [x] Rebased/mergable
- [ ] Tests pass

### DIFF
--- a/charts/telegraf-ds/Chart.yaml
+++ b/charts/telegraf-ds/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: telegraf-ds
-version: 1.0.26
+version: 1.0.27
 appVersion: 1.21.1
 deprecated: false
 description: Telegraf is an agent written in Go for collecting, processing, aggregating, and writing metrics.
--- a/charts/telegraf-ds/templates/configmap.yaml
+++ b/charts/telegraf-ds/templates/configmap.yaml
@@ -39,6 +39,6 @@ data:
 
     [[inputs.kubernetes]]
     url = "https://$HOSTIP:10250"
-    bearer_token = "/run/secrets/kubernetes.io/serviceaccount/token"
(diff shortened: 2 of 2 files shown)


vzabawski:
This is a duplicate of https://github.com/influxdata/helm-charts/pull/311. It was too old and behind commit history, so I had to create a new one.

timhallinflux:
@vzabawski -- check the linter results.

vzabawski:
Hi @timhallinflux. I can fix these errors, but it will be a change not related to the fix introduced with my PR. I used chart version from master branch that already contains those linting errors.
https://github.com/influxdata/helm-charts/blob/master/charts/telegraf-ds/Chart.yaml#L4
https://github.com/influxdata/helm-charts/blob/master/charts/telegraf-ds/templates/role.yaml#L6
https://github.com/influxdata/helm-charts/blob/master/charts/telegraf-ds/templates/rolebinding.yaml#L5

I'm okay with fixing those linting errors since it doesn't look like a huge amount of work, but should it be an additional commit in this pull request, or should I create a separate pull request?

charts/telegraf-ds/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: telegraf-ds
-version: 1.0.22
-appVersion: 1.19
+version: 1.0.23
+appVersion: "1.19"
```
alespour, line 4: Could you rebase and increment `version` (with `appVersion` unquoted) please? Thanks.

alespour (changes requested):

charts/telegraf-ds/Chart.yaml
vzabawski: Sure, no problem.

alespour (approved):

ivankudibal (approved):

Issue #362, opened 2021-09-23
[telegraf] Load to token from env


lpaulusLohr:
Hey trying to load my token from K8S secrets to environnement.

```
- influxdb_v2:
        urls:
          - "http://XXXXXXXXXXX:8086"
        token: "$TOKEN"
        organization: "XXX"
        bucket: "X"
```

Syntax error when Telegraf load config.

```
2021-09-23T07:15:56Z I! Starting Telegraf 1.19.3 
2021-09-23T07:15:56Z I! Using config file: /etc/telegraf/telegraf.conf 
2021-09-23T07:15:56Z E! [telegraf] Error running agent: Error loading config file /etc/telegraf/telegraf.conf: Error parsing data: line 30: invalid TOML syntax
```

The K8S Secret 
```
apiVersion: v1
kind: Secret
metadata:
  name: telegraf-tokens
type: Opaque
data:
  TOKEN: "XXXXXXXXXXXXXXXXXX"
```
I tried:

> `"${TOKEN}"`
> `"$TOKEN"`
> `"\"$TOKEN\""`

lupaulus:
@timhallinflux

lupaulus:
@sjwang90 No body ?

sjwang90:
@lupaulus What does your telegraf.conf look like? Specifically line 30?

lupaulus:
Hi,
I extracted the telegraf conf from my helm Chart.

The Line 30 represent the ´´´token´´´ input, because I use the $ in the Helm chart, he parse it as a String but I need the env value

alespour:
```
        token: "${TOKEN}"
```

produces valid telegraf config line for getting token from environment variable.

The k8s secret should set plain token value without quotes ie.

```yaml
data:
  TOKEN: XXXXXXXXXXXXXXXXXX
```

alespour:
@lupaulus Could you please try the suggestion ^^^.

lpaulusLohr:
Working @alespour

PR #364, opened 2021-10-06
 Add regex for outputs as env variable

lpaulusLohr:
Update _helpers.tpl

Need to be tested


lpaulusLohr:
@timhallinflux Linked to issue #362

PR #365, opened 2021-10-12
chore: update telegraf version

sjwang90:


timhallinflux:
I think you need to bump /charts/telegraf/values.yaml to 1.20-alpine in line 6?

PR #368, opened 2021-10-15
Update to InfluxDB Enterprise 1.9.5

timhallinflux:


rawkode (commented):
Moving this apiVersion to v2 will disallow Helm v1 users; if that's intentional?

rawkode (dismissed):

timhallinflux:
@rawkode -- Error:  Chart.yaml: chart type is not valid in apiVersion 'v1'. It is valid in apiVersion 'v2'
I think that's why I switched it to v2.

rawkode (dismissed):

idclark (dismissed):
LGTM, but not sure how to fix the shared secret for testing.

timhallinflux:
@alespour got further... but still some issues. See details in the check.

alespour:
The issue now is that shared key secret is created in `default` namespace, but `ct install` deploys pods to a generated namespace (`influxdb-enterprise-ty4rjto6yu`), so the secret is not visible to them. A simple fix for that is to tell `ct` to use default namespace ie.

https://github.com/influxdata/helm-charts/blob/df36a15316ae1bc50af74608acffcf60a3720add/.github/workflows/helm-charts-test.yaml#L52
should be `run: ct install --namespace=default`. I wonder how come it worked before, maybe it's changed in some later version or chart testing tools.

The next issue we hit in our fork was with the license key. I'm not sure how it is supposed to work in the workflow, but in the end I refactored the chart and the workflow to allow the license key to be passed to pod's environment via a secret. It is in the [a27f8a8](https://github.com/bonitoo-io/helm-charts/pull/8/commits/a27f8a8a48ba1ce2741738b2a4a6e2c63bd7f131) commit in out fork.

timhallinflux:
@rawkode -- any comments on this?

alespour:
Now the license key is not found. The secret `influxdb-license` is commented out in `example-resources.yaml` for obvious reason, so I think it has to be created in the workflow. A possible way to continue could then be

1. in `.github/workflows/helm-charts-test.yaml`, add the following after `kubectl apply -f ./charts/influxdb-enterprise/example-resources.yaml` 
```yaml
          ...
          kubectl create secret generic influxdb-license --from-literal=json=${INFLUXDB_ENTERPRISE_LICENSE_KEY}
        env:
          INFLUXDB_ENTERPRISE_LICENSE_KEY: "${{ secrets.INFLUXDB_ENTERPRISE_LICENSE_KEY }}"
        if: steps.list-changed.outputs.changed == 'true'
```
2. create folder `ci` in `charts/enterprise-influxdb`, copy `values.yaml` there as `license-values.yaml` (for example), and configure license to use a secret like
```yaml
license:
  # You can put your license key here for testing this chart out,
  # but we STRONGLY recommend using a license file stored in a secret
  # when you ship to production.
  # key: ""
  secret:
    name: influxdb-license
    key: json
```
Chart testing will automatically use `ci/***-values.yaml` file(s) instead of `values.yaml in the root folder.

alespour:
Now the nodes can't find license file `/var/run/secrets/influxdb/license.json`. As I understand the chart, when the license key is not set (in the `values.yaml`), influxdb nodes are configured (via `data|meta-configmap.yaml`) to read the license from that file. There should be a kubernetes secret with license file content (it mounted as `/var/run/secrets/influxdb/license.json` later on). But, the secret `influxdb-license` is commented out in `example-resources.yaml`, for obvious reasons.

I do not know how it was done before in the CI test, but I think it could done be like that the content of the license file is stored in the repository secret, and then the kubernetes secret created in the workflow, like

1. in `.github/workflows/helm-charts-test.yaml`, add the following after the `kubectl apply -f ./charts/influxdb-enterprise/example-resources.yaml`  line:
```yaml
          ...
          kubectl create secret generic influxdb-license --from-literal=json=${INFLUXDB_ENTERPRISE_LICENSE_JSON}
        env:
          INFLUXDB_ENTERPRISE_LICENSE_JSON: "${{ secrets.INFLUXDB_ENTERPRISE_LICENSE_JSON }}"
        if: steps.list-changed.outputs.changed == 'true'
```
2. create folder `ci` in `charts/enterprise-influxdb`, copy `values.yaml` there as `license-values.yaml` (for example), and set the license secret and key like
```yaml
license:
  # You can put your license key here for testing this chart out,
  # but we STRONGLY recommend using a license file stored in a secret
  # when you ship to production.
  # key: ""
  secret:
    name: influxdb-license
    key: json
```

The content of the key from the secret is mounted to `/var/run/secrets/influxdb/license.json` in `data|meta-statefulset.yaml` descriptor and the influxdb nodes should be able to read it then.

---

_I could not test it_ because we don't have license file, just the key. So I have modified the chart and the workflow ([commit a27f8a8](https://github.com/bonitoo-io/helm-charts/pull/8/commits/a27f8a8a48ba1ce2741738b2a4a6e2c63bd7f131)) to allow passing the license key to nodes vie env var.

PR #386, opened 2021-11-27
Update ingress to v1 from v1beta1 for Influxdb2.

gregnrobinson:
Not sure if this was by design, but to get the code working I had to copy the config from `influxdbv1` to `influxdbv2` which had the v1 networking config. `influxdbv2` has `v1beta1` at least on the master branch that is. I just copied it over as is and it has worked since I did that. I am sure other folks will run into the problem if not already.

Version 1 has the new config and version 2 has the old one.

### DIFF
--- a/charts/influxdb2/templates/ingress.yaml
+++ b/charts/influxdb2/templates/ingress.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.ingress.enabled -}}
-{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
 apiVersion: networking.k8s.io/v1
 {{- else }}
 apiVersion: networking.k8s.io/v1beta1
--- a/charts/influxdb2/templates/ingress.yaml
+++ b/charts/influxdb2/templates/ingress.yaml
@@ -1,30 +1,42 @@
 {{- if .Values.ingress.enabled -}}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
+apiVersion: networking.k8s.io/v1
+{{- else }}
 apiVersion: networking.k8s.io/v1beta1
+{{- end }}
 kind: Ingress
 metadata:
   name: {{ include "influxdb.fullname" . }}
   labels:
     {{- include "influxdb.labels" . | nindent 4 }}
-  {{- with .Values.ingress.annotations }}
   annotations:
-    {{- toYaml . | nindent 4 }}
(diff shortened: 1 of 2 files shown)


naimo84:
Would love it, if this PR is merged soon 👍

charts/influxdb2/templates/ingress.yaml
```diff
@@ -1,30 +1,42 @@
 {{- if .Values.ingress.enabled -}}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
```
alespour, line 2: Could you please change the condition to check for Ingress object in v1 since some k8s versions have v1 API but that does include Ingress (it's since 1.19)? Thank you.

`{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}`

charts/influxdb2/templates/ingress.yaml
```diff
@@ -1,30 +1,42 @@
...
+{{- end }}
+{{- if .Values.ingress.className }}
+  ingressClassName: {{ .Values.ingress.className }}
 {{- end }}
   rules:
   - host: {{ .Values.ingress.hostname }}
     http:
       paths:
       - path: {{ .Values.ingress.path }}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
```
alespour, line 29: dtto

charts/influxdb2/templates/ingress.yaml
```diff
@@ -1,30 +1,42 @@
...
+        pathType: Prefix
+{{- end }}
         backend:
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
```
alespour, line 33: dtto

alespour (changes requested):
Could you please increment `version` in `Chart.yaml`.

charts/influxdb2/templates/ingress.yaml
gregnrobinson: added.

charts/influxdb2/templates/ingress.yaml
gregnrobinson: added.

charts/influxdb2/templates/ingress.yaml
gregnrobinson: added.

gregnrobinson:
I wasn't sure exactly what the increment should be so I updated it from `2.0.2` to `2.0.3`

alespour (approved):
Thank you!

ivankudibal (approved):

PR #388, opened 2021-11-30
Fix ingress capability check in influxdb

PaulFarver:
Ensure that the Ingress object is present in networking.k8s.io/v1, before rendering that version of the object.

Some versions of k8s (ie. 1.16), have the v1 api, but it does not include the Ingress, which causes the install to fail, when the ingress is enabled.

The helm documentation states that it is possible to check for an objects presence in an api group:
https://helm.sh/docs/chart_template_guide/builtin_objects/

### DIFF
--- a/charts/influxdb/Chart.yaml
+++ b/charts/influxdb/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: influxdb
-version: 4.10.1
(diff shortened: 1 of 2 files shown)


alespour (approved):
Thank you!

ivankudibal (approved):

PR #391, opened 2021-12-04
Use port name for backend selector in influxdb2

siku2:
In #386 the service port was hard coded to port 8086. Unfortunately, that's not necessarily the port that the service exposes. In fact, it's not even the default!
The default values.yaml config uses port 80 for the service. Since the ingress specifies the service port, not the container port, the ingress specification doesn't work as expected.

### DIFF
--- a/charts/influxdb2/templates/ingress.yaml
+++ b/charts/influxdb2/templates/ingress.yaml
@@ -37,6 +37,6 @@ spec:
               name: {{ .Values.service.portName }}
 {{- else }}
           serviceName: {{ include "influxdb.fullname" . }}
-          servicePort: 8086
+          servicePort: {{ .Values.service.portName }}
 {{- end }}
 {{- end -}}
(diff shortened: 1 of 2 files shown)


alespour (changes requested):
Thank you for the contribution. Could you please increment the `version` in`Chart.yaml` to `2.0.4`?

charts/influxdb2/templates/ingress.yaml
```diff
@@ -34,7 +34,7 @@ spec:
           service:
             name: {{ include "influxdb.fullname" . }}
             port:
-              number: 8086
+              name: {{ .Values.service.portName }}
 {{- else }}
           serviceName: {{ include "influxdb.fullname" . }}
           servicePort: 8086
```
alespour, line 40: Could you please fix it here too. Thank you.

siku2: I wasn't familiar with the v1beta1 spec, so I assumed it was correct there. I managed to track down the spec though and it also supports the service port name so I changed it there too.

alespour (approved):

ivankudibal (approved):

PR #397, opened 2021-12-13
chore: update chart releaser action and Helm installation

alespour:
- releaser 1.0.0-rc.1 -> 1.2.1
- Helm installation via action (the same as in tests)

### DIFF
--- a/.github/workflows/helm-charts-release.yaml
+++ b/.github/workflows/helm-charts-release.yaml
@@ -41,6 +41,6 @@ jobs:
           ./get_helm.sh
 
       - name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1.0.0-rc.1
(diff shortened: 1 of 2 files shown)


alespour:
Tested in our fork, seems to work just fine.

timhallinflux:
@rawkode -- can you have a quick look at this?

alespour:
So there is an option to skip releasing a chart is release already exist. This should fix problems _in the future_. Added in [5bbb916](https://github.com/influxdata/helm-charts/pull/397/commits/5bbb916caaa722a3e1d24d7585c56e260db5df69) commit.

Also updated Helm in the workflow to recent v3.6.3 (latest release is 3.7.2 patch release so 3.6.x should be super-stable I assume).

timhallinflux (approved):

Issue #164, opened 2020-08-22
[telegraf] input port not exposed in Service


aureq:
Hi

Telegraf supports ``[[inputs.syslog]]`` however, the plugin port is not exposed in the telegraf service during deployment.

The current generated output:
```yaml
# Source: telegraf/templates/service.yaml
apiVersion: v1
kind: Service
metadata:
  name: telegraf
  labels:
    helm.sh/chart: telegraf-1.7.21
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: telegraf
    app.kubernetes.io/instance: telegraf
spec:
  type: ClusterIP
  ports:
  - port: 8888
    targetPort: 8888
    name: "health"
  selector:
    app.kubernetes.io/name: telegraf
    app.kubernetes.io/instance: telegraf
```

If the syslog input plugin is enabled, then it would be good to have

Desired output
```yaml
# Source: telegraf/templates/service.yaml
apiVersion: v1
kind: Service
metadata:
  name: telegraf
  labels:
    helm.sh/chart: telegraf-1.7.21
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: telegraf
    app.kubernetes.io/instance: telegraf
spec:
  type: ClusterIP
  ports:
  - port: 8888
    targetPort: 8888
    name: "health"
  - port: 514
    targetPort: 514
    protocol: UDP
    name: "syslog"
  selector:
    app.kubernetes.io/name: telegraf
    app.kubernetes.io/instance: telegraf
```

Thanks.

russorat:
@aureq thanks! would this be a fix you'd like to contribute?

aureq:
Hi @russorat 

Happy to try that, however I have limited bandwidth right now.

alespour:
Closed via #401

PR #401, opened 2022-01-01
expose syslog port

aSauerwein:
automatically expose syslog port as defined in config.inputs
this fixes issue #164 
```
- [ ] CHANGELOG.md updated
- [ ] Rebased/mergable
- [ ] Tests pass
```

### DIFF
--- a/charts/telegraf/templates/service.yaml
+++ b/charts/telegraf/templates/service.yaml
@@ -63,6 +63,12 @@ spec:
     targetPort: {{ trimPrefix ":" $value.service_address | int64 }}
     name: "webhooks"
     {{- end }}
+    {{- if eq $key "syslog" }}
+  - port: {{ regexFind "[0-9]+$" $value.server  | int64 }}
+    targetPort: {{ regexFind "[0-9]+$" $value.server  | int64 }}
+    protocol: {{ upper (substr 0 3 $value.server) }}
+    name: "syslog"
+    {{- end }}  
     {{- if eq $key "socket_listener" }}
     {{- if or (hasPrefix "udp" $value.service_address) (hasPrefix "tcp" $value.service_address) }}
   - port: {{ regexFind "[0-9]+$" $value.service_address  | int64 }}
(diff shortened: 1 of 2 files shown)


charts/telegraf/templates/service.yaml
```diff
@@ -62,6 +62,12 @@ spec:
     targetPort: {{ trimPrefix ":" $value.service_address | int64 }}
     name: "webhooks"
     {{- end }}
+    {{- if eq $key "syslog" }}
+  - port: {{ regexFind "[0-9]+$" $value.server  | int64 }}
```
alespour, line 66: To make it safe I'd suggest to add protocol condition check (with unix socket server address the rendered port spec is invalid), eg.

```suggestion
    {{- if eq $key "syslog" }}
      {{- if regexMatch "^(tcp|udp).*" $value.server }}
  - port: {{ regexFind "[0-9]+$" $value.server  | int64 }}
```

charts/telegraf/templates/service.yaml
```diff
@@ -62,6 +62,12 @@ spec:
...
+    targetPort: {{ regexFind "[0-9]+$" $value.server  | int64 }}
+    protocol: {{ upper (substr 0 3 $value.server) }}
+    name: "syslog"
+    {{- end }}
```
alespour, line 70:
```suggestion
      {{- end }}
    {{- end }}
```

alespour (changes requested):

alespour:
Closes #164

alespour (approved):

tomklapka:
Does it make sense to you @alespour to improve also other Telegraf charts (telegraf-ds, telegraf-operator) ?

timhallinflux (approved):

alespour:
> Does it make sense to you @alespour to improve also other Telegraf charts (telegraf-ds, telegraf-operator) ?

It is not applicable - `telegraf-ds` has hardcoded inputs without syslog and `telegraf-ds` has no input plugins

PR #408, opened 2022-01-12
Fix probes and internal metrics configuration

### LINKED ISSUE #264: [telegraf] Can't disable internal input
The internal input is hardcoded in and there's no conditional around it like for the health output.

https://github.com/influxdata/helm-charts/blob/master/charts/telegraf/templates/configmap.yaml#L24-L25

Seems strange to me to not have an option to disable it?

### LINKED ISSUE #406: Readiness/Liveness probes not configurable properly
When using the metrics configuration like below, i would expect that a liveness/readyness probe is configured. But this is not the case.

Example values.yaml:

```
metrics:
  health:
    enabled: true
    threshold: "5000.0"
```

This setup creates an ``outputs.health`` configuration and enables the required port in the service object.

But the readiness/liveness probes are only created when there is an ``outputs.health`` plugin configured in the config section. 

The relevant part from deployment.yaml is below

```
{{- range $objectKey, $objectValue := .Values.config.outputs }}
        {{- range $key, $value := . -}}
        {{- $tp := typeOf $value -}}
        {{- if eq $key "health" }}
        livenessProbe:
          httpGet:
            path: /
            port: {{ trimPrefix "http://:" $value.service_address | int64 }}
        readinessProbe:
          httpGet:
            path: /
            port: {{ trimPrefix "http://:" $value.service_address | int64 }}
        {{- end }}
        {{- end }}
        {{- end }}
```

Whilst it is possible to configure such an output anoter service address (not port 8888) must be used which renders the outputs.health created in the metrics section useless.

Did i get somwthing wrong?

alespour:
- closes #406 
  - adds test files for default values test and metrics health enabled
- closes #264 
  -  `internal` plugin is enabled by default for backward compatibility, can be disabled with `metrics.internal.enabled: false`
- updates Telegraf to 1.21.2

### DIFF
--- a/charts/telegraf/Chart.yaml
+++ b/charts/telegraf/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: telegraf
-version: 1.8.10
-appVersion: 1.21.1
+version: 1.8.11
+appVersion: 1.21.2
 deprecated: false
 description: Telegraf is an agent written in Go for collecting, processing, aggregating, and writing metrics.
 keywords:
--- a/charts/telegraf/ci/default-values.yaml
+++ b/charts/telegraf/ci/default-values.yaml
@@ -1,0 +1,1 @@
+# Default values only, no overrides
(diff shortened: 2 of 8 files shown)


tomklapka (approved):
Looks good

ivankudibal (approved):

PR #409, opened 2022-01-23
Telegraf-DS: parametrize the pod's SecurityContext

eacherkan-aternity:
This PR exposes the Telegraf pod's SecurityContext as a user-configured value, allowing the user to provide the `runAsUser`/`runAsGroup` parameters in the Telegraf-DS chart. (The SecurityContext is already exposed in similar way in the Telegraf and Telegraf-Operator charts.) Since Telegraf 1.20.3, providing a non-default `runAsGroup` value is crucial to being able to use the Docker input plugin (see [Docker: Run Telegraf as non-root](https://www.influxdata.com/blog/docker-run-telegraf-as-non-root/)).

I tested the fix by running `helm template` and examining the output.

- [ ] CHANGELOG.md updated
- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/charts/telegraf-ds/templates/daemonset.yaml
+++ b/charts/telegraf-ds/templates/daemonset.yaml
@@ -83,3 +83,5 @@ spec:
       dnsConfig:
 {{ toYaml . | indent 8 }}
       {{- end }}
+      securityContext:
+{{ toYaml .Values.podSecurityContext | indent 8 }}
--- a/charts/telegraf-ds/templates/daemonset.yaml
+++ b/charts/telegraf-ds/templates/daemonset.yaml
@@ -83,5 +83,7 @@ spec:
       dnsConfig:
 {{ toYaml . | indent 8 }}
       {{- end }}
+      {{- with .Values.podSecurityContext }}
       securityContext:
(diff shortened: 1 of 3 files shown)


charts/telegraf-ds/templates/daemonset.yaml
```diff
@@ -83,3 +83,5 @@ spec:
       dnsConfig:
 {{ toYaml . | indent 8 }}
       {{- end }}
+      securityContext:
+{{ toYaml .Values.podSecurityContext | indent 8 }}
```
alespour, line 87: Using `with` prevents rendering empty `podSecurityContext` dictionary.

```suggestion
      {{- with .Values.podSecurityContext }}
      securityContext:
{{ toYaml . | indent 8 }}
      {{- end }}
```

alespour (changes requested):
Also, could you please bump chart version? Thank you.

eacherkan-aternity:
Thanks @alespour, I merged the change to `with` + bumped the chart version to 1.0.28.

alespour (approved):
Thank you for the contribution!

tomklapka (approved):

PR #419, opened 2022-02-08
Get the KAPACITOR_SLACK_URL from a secret

afausti:
- Avoid publishing publicly the Slack handler URL in the values.yaml. Obtain that from the Kapacitor `existingSecret` instead.

### DIFF
--- a/charts/kapacitor/templates/deployment.yaml
+++ b/charts/kapacitor/templates/deployment.yaml
@@ -49,6 +49,12 @@ spec:
             secretKeyRef:
               key: influxdb-password
               name: {{ .Values.existingSecret }}
+        - name: KAPACITOR_SLACK_URL
+          valueFrom:
+            secretKeyRef:
+              key: kapacitor-slack-url
+              name: {{ .Values.existingSecret }}
+              optional: true
         {{- end }}
         ports:
         - containerPort: 9092
--- a/charts/kapacitor/Chart.yaml
+++ b/charts/kapacitor/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kapacitor
-version: 1.4.2
+version: 1.4.3
 appVersion: 1.6.3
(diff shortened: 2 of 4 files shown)


charts/kapacitor/templates/deployment.yaml
```diff
@@ -49,6 +49,12 @@ spec:
             secretKeyRef:
               key: influxdb-password
               name: {{ .Values.existingSecret }}
+        - name: KAPACITOR_SLACK_URL
+          valueFrom:
+            secretKeyRef:
+              key: kapacitor-slack-url
+              name: {{ .Values.existingSecret }}
+              optional: true
```
alespour, line 57: Just a thought - it the `kapacitor-` prefix of the key needed to distinguish it from another Slack URL? I mean influxdb-related keys do not have such prefix.

And second, given there can be multiple slack clients, then, similarly to influxdb env vars injection, having `_0_` index in the env var name would seem like using the same [correct naming style](https://docs.influxdata.com/kapacitor/v1.6/administration/configuration/#map-configuration-properties-to-environment-variables), I think, ie something like:

```suggestion
        - name: KAPACITOR_SLACK_0_URL
          valueFrom:
            secretKeyRef:
              key: slack-url
              name: {{ .Values.existingSecret }}
              optional: true
```
?

alespour (changes requested):

charts/kapacitor/templates/deployment.yaml
afausti: Hey, @alespour thanks for looking into this. Good point I'll remove the `kapacitor-` prefix to uniformize the secret keys.  

After reading the [naming style ](https://docs.influxdata.com/kapacitor/v1.6/administration/configuration/#map-configuration-properties-to-environment-variables) for the environment variables I thought `KAPACITOR_SLACK_URL` would match the path in the configuration file, but I missed the fact that we can have multiple Slack clients. Let me test that and commit a fix.

PR #430, opened 2022-03-09
[influxdb2] add service account

### LINKED ISSUE #322: [influxdb2] helm chart creates a new admin-password/token even if bolt file is present
When a bolt file is present (eg. persistence is used), the initialisation is skipped in the docker entrypoint. However the helm chart will generate a new token and password even in that case. This is confusing considering that the chart will tell you at the end of the deploy a new (useless) password even though you still need to use the old one.

### LINKED ISSUE #428: [influxdb2] Set automountServiceAccountToken
I ran kubeaudit against my k3s cluster in which I deployed the influxdb2 chart. It reports the following:
```
-- [error] AutomountServiceAccountTokenTrueAndDefaultSA
   Message: Default service account with token mounted. automountServiceAccountToken should be set to 'false' on either the ServiceAccount or on the PodSpec or a non-default service account should be used.
```
However I cannot find an option to specify this in the values.yaml file of the chart. I also believe automountServiceAccountToken should be set to false by default in the chart.

Thanks for your time and help.

helm list output:
```
NAME     	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART                  	APP VERSION
influxdb2	default  	6       	2022-03-03 21:15:04.043523702 +0000 UTC	deployed	influxdb2-2.0.9        	2.1.1
```

alespour:
- adds service account
Closes #428
- adds note about admin password persistence
Closes #322

### DIFF
--- a/charts/influxdb2/Chart.yaml
+++ b/charts/influxdb2/Chart.yaml
@@ -4,7 +4,7 @@ name: influxdb2
 description: A Helm chart for InfluxDB v2
 home: https://www.influxdata.com/products/influxdb/
 type: application
-version: 2.0.9
+version: 2.0.10
 maintainers:
   - name: rawkode
     email: rawkode@influxdata.com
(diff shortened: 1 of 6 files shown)


tomklapka (approved):

timhallinflux:
still in draft?

alespour:
@timhallinflux I forgot to flip it, sorry

ivankudibal (approved):

PR #440, opened 2022-03-15
update SA name for telegraf roles

### LINKED ISSUE #323: [telegraf-ds] Use serviceAccountName in ClusterRole
If I try and use an existing service account for the `telegraf-ds` chart - ie.,

```
  serviceAccount:
    create: false
    name: my-sa
```

The resulting `ClusterRoleBinding` seems to end up with the incorrect SA name - eg,

```
  - kind: ServiceAccount
    name: <value of telegraf.fullname>
```

I tested this diff and it seemed to correct the SA name:

```diff
--- a/charts/telegraf-ds/templates/rolebinding.yaml
+++ b/charts/telegraf-ds/templates/rolebinding.yaml
@@ -7,10 +7,10 @@ metadata:
     {{- include "telegraf.labels" . | nindent 4 }}
 subjects:
   - kind: ServiceAccount
-    name: {{ include "telegraf.fullname" . }}
+    name: {{ include "telegraf.serviceAccountName" . }}
     namespace: {{ .Release.Namespace }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
```

It also stopped these errors reported by telegraf (associated with the binding not being applied):

```
E! [inputs.kubernetes] Error in plugin: https://nn.nn.nn.nn:10250/stats/summary returned HTTP status 403 Forbidden
```

If you think this is a valid issue and agree with the fix I'd be happy to make a PR for it.

Thanks

dougsc:
Update `telegraf` bindings to use the correct SA template. 

This MR closes #323


timhallinflux (approved):

timhallinflux (changes requested):
need to bump the chart versions on each

alespour:
The issue with the failing `telegraf-operator` check now is that unlike in other charts, where ServiceAccount descriptor is wrapped in 
```yaml
{{- if .Values.serviceAccount.create }}
....
{{- end }}
```
condition, it is **not** in the telegraf-operator chart, where ServiceAccount is always created. However with no name set in values, it default to "default" which already exists in the cluster by default, hence the error. I assume it is ok to use the same `if-end` in the `serviceaccount.yaml` of the telegraf-operator chart too.

wojciechka:
> I assume it would be ok and appropriate to use the same `if-end` in the `serviceaccount.yaml` of the telegraf-operator chart too. Or should telegraf-operator always create service account, @wojciechka ?

`telegraf-operator` always needs a service account as it needs to have specific cluster roles assigned to it. It should not be otpional.

Can you please remove telegraf-operator changes from this PR and open an issue (here or in [influxdata/telegraf-operator](https://github.com/influxdata/telegraf-operator)) to address issues with service account name being empty or not set in that chart? I'm happy to check and don't want to block you from fixing other charts.

alespour (approved):
Thank you for the contribution, this was a good catch.

tomklapka (approved):

PR #450, opened 2022-03-23
[influxdb] set default user without job

### LINKED ISSUE #85: initScripts is incompatible with http-enabled: true
Figuring out how to both enable InfluxDB authentication and use an init script was a cumbersome process and I'm not happy with the solution I found.

Readme tells (https://github.com/influxdata/helm-charts/tree/master/charts/influxdb#starting-with-authentication):
>  To handle this setup on startup, a job can be enabled in values.yaml by setting .Values.setDefaultUser.enabled to true.

SetDefaultUser is an asynchronous process run in a dedicated container. Init scripts are run by container entrypoint. If admin user is created after init script is executed, commands are rejected because there is no admin user. If admin user is created before init script is executed, commands are rejected because init script doesn't provide, nor know, the right credentials.

It also not possible to create an admin user from the init script. 

I figured out it was possible to set INFLUXDB_ADMIN_(USER|PASSWORD) environment variables. It this case, entrypoint creates the admin user first and then provides the proper credentials to subsequent commands. 

This leads me to ask two questions:

 - Shouldn't set-auth post-install hook be superseded by another mechanism ? 
 - Is there is way to use secrets to set INFLUXDB_ADMIN_PASSWORD ?

### LINKED ISSUE #292: Influxdb Charts set-auth job needs nodeSelector
When I install the Influxdb chart with a `nodeSelector`, the main pod chooses correct nodes. However the set-auth job does not have a `nodeSelector` and the image is not a multi-arch build. 

Here is the template I speak of: https://github.com/influxdata/helm-charts/blob/master/charts/influxdb/templates/post-install-set-auth.yaml

For whatever reason the set-auth job really wants to run on my ARM nodes and therefore gets the usual `exec user format error` which simply means wrong architecture. 

I suggest to simply use the existing `nodeSelector` value from the main pod to reduce configuration. This way the all the pods will run on the nodes I select. Otherwise just add a `nodeSelector` value for the set auth job.

### LINKED ISSUE #305: influxdb setDefaultUser not working on arm platform
I ran into this issue when trying to deploy influxdb on arm64 platform. While influxdb image works just fine, the appropriate/curl image it references to set user with password does not have an arm64 version and fails. Changing image to curlimages/curl:latest seems to fix the issue.

alespour:
Supersedes and closes #447 therefore closes #292 closes #305 closes #85 (for good)

`init-influxdb.sh` in the docker image supports default user initialization via environment variables so there is no need to use a job. It solves chicken-egg problem when http authentication is enabled and init script is used too described in #85 more user friendly and according to docs than presented in [85#issuecomment-1075283672](https://github.com/influxdata/helm-charts/issues/85#issuecomment-1075283672).


tomklapka (approved):

timhallinflux (approved):

brsolomon-deloitte:
How does a major change to values.yaml schema constitute a patch version chart release? 🤔

PR #463, opened 2022-04-18
[telegraf-ds] - Feature request add custom annotations and labels to daemonSet and daemonSet spec template 

ariverflows:
## Feature Request
- Added additionalLabels value to default values file to allow users to add labels to daemonset
- Added podLabels value to default values file to allow users to add pod labels to daemonset spec template
- Added podAnnotations value to default values file to allow users to add pod annotations to daemonset spec template

## Testing
- Validated locally with helm deployment

- [ x ] Tests pass
- [ x ] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

Signed-off-by: AdrianRivera <Adrian.Rivera@target.com>

### DIFF
--- a/charts/telegraf-ds/Chart.yaml
+++ b/charts/telegraf-ds/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: telegraf-ds
-version: 1.0.34
-appVersion: 1.22.0
+version: 1.0.35
+appVersion: 1.23.0
 deprecated: false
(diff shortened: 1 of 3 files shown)


ranyhb:
+1 — we need exactly this. `telegraf-ds` exposes `podLabels` and pod annotations, but the DaemonSet object's own metadata takes only `name` and `labels`, so there is no way to put an annotation on the DaemonSet itself.

Our case is a Pulumi deployment that needs `pulumi.com/skipAwait` on the DaemonSet; today we post-process the rendered manifest to get it. Happy to help test or review.

PR #473, opened 2022-05-19
Bugfix #472: Make default value for influxURL available and mandatory

### LINKED ISSUE #472: Kapacitor: Non-existent default value for  .Values.influxURL prevents deployment from being rendered
The deployment for Kapacitor  is only rendered if `.Values.influxURL` is set:
https://github.com/influxdata/helm-charts/blob/bae0041c452343c91d6123fca199471a8e1aaf95/charts/kapacitor/templates/deployment.yaml#L1-L2

That value is documented as having a default value of `http://influxdb-influxdb.tick:8086`:

https://github.com/influxdata/helm-charts/blob/bae0041c452343c91d6123fca199471a8e1aaf95/charts/kapacitor/README.md?plain=1#L64

However, it is not set in the default `values.yaml`:
https://github.com/influxdata/helm-charts/blob/bae0041c452343c91d6123fca199471a8e1aaf95/charts/kapacitor/values.yaml

This causes the chart to render without deployment if the default values are used.

mwmahlberg:
This PR fixes #472.

Tests run:
- Ran helm template with new values.yaml and the chart rendered correctly
- Ran helm without influxURL set and an error was thrown 

- [ ] CHANGELOG.md updated
- [ ] Rebased/mergable
- [ ] Tests pass

### DIFF
--- a/charts/kapacitor/templates/deployment.yaml
+++ b/charts/kapacitor/templates/deployment.yaml
@@ -1,5 +1,3 @@
-{{- $bl := empty .Values.influxURL }}
-{{- if not $bl }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -33,7 +31,7 @@ spec:
         - name: KAPACITOR_HOSTNAME
           value: {{ template "kapacitor.fullname" . }}.{{ .Release.Namespace }}
         - name: KAPACITOR_INFLUXDB_0_URLS_0
-          value: {{ .Values.influxURL }}
+          value: {{ .Values.influxURL | required "Value influxURL not set!"}}
         {{- range $key, $val := .Values.envVars }}
         - name: {{ $key }}
           value: {{ $val | quote }}
@@ -112,4 +110,3 @@ spec:
         - name: sc-sideload-volume
           emptyDir: {}
       {{- end -}}
-{{- end }}
--- a/charts/kapacitor/values.yaml
+++ b/charts/kapacitor/values.yaml
@@ -10,6 +10,8 @@ image:
 # strategy:
 #   type: Recreate
 
+strategy: {}
+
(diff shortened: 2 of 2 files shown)


alespour:
Shouldn't installation just fail when `influxURL` is not set, instead of using nonexistent default URL? Kapacitor won't work properly anyway.

mwmahlberg:
@alespour It does so with this -- it literally prints `Value influxURL not set!` if the value is not set -- a much more useful way for the user, imho. See

https://github.com/mwmahlberg/helm-charts-1/blob/ea0b855a878c0943fc17b5926a13d264b2795b72/charts/kapacitor/templates/deployment.yaml#L34

for details.

mwmahlberg:
@alespour Are there any questions left or something that needs to be done?

alespour:
> It does so with #473 - it literally prints `Value influxURL not set!` if the value is not set. This imho is a much 

Yes, but with this PR `influxURL` the value is always set (there is default in `values.yaml`), so it will **not** make the user to set it to some meaningful value because it will never fail and print the warning.

mwmahlberg:
@alespour I sure could remove the default value. My intention was to got for a two pronged approach: For one to make sure the default install will actually install and in case influxURL is not set to return a more meaningful error.

I see two approaches here: The default value could be set to an empty string:

```yaml
influxURL: ""
```

This would [evaluate to false](https://go.dev/play/p/QLc6kAVAVM1) and hence trigger the `required` part of the expression.
However, the value would be present in values.yaml and there would be a meaningful error message, even before an install is attempted.

The other approach could be to simply set a default value, however this could be a bit irritating when done on its own.

mwmahlberg:
@alespour So, which way should I modify the PR, or should I close it?

tomklapka (approved):

Issue #433, opened 2022-03-10
[Telegraf] ConfigMap telegraf Yaml => Toml error


lpaulusLohr:
Hi,

I have an problem with ConfigMap Yaml to Toml parsing : 

I want this  : 
```
    [[inputs.mqtt_consumer.json_v2.object]]
        path = "@this"
        tags = ["device","data.type","owner"] 
```

But the helm parsing template render it like : 
```
    [[inputs.mqtt_consumer.json_v2.object]]
        path = "@this"
        tags = "[device data.type owner]"
```

My Yaml :

```
          object:
            path: "@this"
            tags:
              - "device"
              - "data.type"
              - "owner"
```

I found the problem is here in the template  *_helpers.tpl*  l.277
```
          [inputs.{{ $input }}.{{ $key }}.{{ $k }}]
                {{- range $foo, $bar := $v }}
            {{ $foo }} = {{ $bar | quote }}
```

alespour:
Could you please try the latest chart with `--set tplVersion=2`.

[README.md#telegraf-configuration](https://github.com/influxdata/helm-charts/blob/master/charts/telegraf/README.md#telegraf-configuration)

alespour:
Should be
```toml
  [[inputs.mqtt_consumer.json_v2]]
    measurement_name_path = "device"
    [[inputs.mqtt_consumer.json_v2.object]]  
      path = "@this"
      tags = [
        "device",
        "data.type",
        "owner"
      ]   
```
for
```yaml
        json_v2:
          - measurement_name_path: "device"
            object:
              - path: "@this"
                tags:
                  - "device"
                  - "data.type"
                  - "owner"
```

lpaulusLohr:
It's a Match 👍 : 

```
json_v2:
  measurement_name_path: "device"
  object:
    path: "@this"
    tags:
      - "device"
      - "data.type"
      - "owner"
 ```

alespour:
I see :)
OK to close the issue?

lpaulusLohr:
I did not pay attention, but there is a problem on the title of the "section", there is only 1 hook instead of 2

```javascript
[[inputs.mqtt_consumer]]    
  data_format = "json_v2"

  topics = [
    "XXXX"
  ]
  [inputs.mqtt_consumer.json_v2]  
    measurement_name_path = "device"
    [inputs.mqtt_consumer.json_v2.object]  
      path = "@this"
      tags = [
        "device",
        "data.type",
        "owner"
      ]
[...]

PR #501, opened 2022-09-14
feat: update telegraf to v1.24.0

powersj:
Creating PR to see if 2 spaces in our automation passes checks

### DIFF
--- a/charts/telegraf-ds/Chart.yaml
+++ b/charts/telegraf-ds/Chart.yaml
@@ -1,22 +1,22 @@
 apiVersion: v1
 name: telegraf-ds
-version: 1.1.1
-appVersion: 1.23.3
+version: 1.1.2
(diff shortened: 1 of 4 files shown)


powersj:
This update is only updating versions, so I would not have expected this failure.

The current CI failure is from the [complex-values](https://github.com/influxdata/helm-charts/blob/master/charts/telegraf/ci/complex-values.yaml) test case:

```
2022-09-14T16:54:48Z E! [telegraf] Error running agent: Error loading config file /etc/telegraf/telegraf.conf: plugin inputs.mqtt_consumer: line 93: configuration specified the fields ["json_v2"], but they weren't used
```

It looks like it is trying to render json_v2 as a field and not a table. This case came out of #433

powersj:
closing in favor of upcoming v1.24.1

PR #510, opened 2022-10-02
Fix #417 Add serviceMonitor template for the telegraf chart

### LINKED ISSUE #417: Missing Feature: Prometheus ServiceMonitor for [[outputs.prometheus_client]] in telegraf chart
We are using the ``[[outputs.prometheus_client]]`` plugin in telegraf. The plugin itself is working as expected. 

From my experience, most installations use the prometheus-operator when deploying prometheus on k8s.
prometheus-operator uses a ``ServiceMonitor`` CRD to automatically detect metric endpoints (here: endpoint exposed by ``[outputs.prometheus_client]]``)

This ServiceMonitor is normally deployed with the application to be monitored. As an example see [ServiceMonitor template in Grafana Chart](https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/servicemonitor.yaml). So it would be desirable to have a ``ServiceMonitor`` template added to the chart.

As a workaround i manually deployed a ServiceMonitor manifest after deploying the telegraf chart.
````
# Prometheus ServiceMonitor for telegraf
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: telegraf
  namespace: monitoring
  labels:
    app: telegraf
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: telegraf
  namespaceSelector:
    matchNames:
    - monitoring
  endpoints:
    - port: prometheus-client
      path: /metrics
````

See also:

- [Prometheus Operator Docs](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md)

Let me note that i did not read your ``Contributing" section yet. If it helps i can provide a suggestion for ServiceMonitor template, just let me know.

### PR DESCRIPTION
**Pull request by Efrat19**
Signed-off-by: Efrat19 <efrat890089@gmail.com>

### DIFF
--- a/charts/telegraf/templates/servicemonitor.yaml
+++ b/charts/telegraf/templates/servicemonitor.yaml
@@ -1,0 +1,22 @@
+{{- if .Values.serviceMonitor.create }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ template "telegraf.serviceMonitorName" . }}
+  labels:
+    {{- include "telegraf.labels" . | nindent 4 }}
+  {{- if .Values.serviceMonitor.annotations }}
+  annotations:
+    {{- toYaml .Values.serviceMonitor.annotations | nindent 4 }}
+  {{- end }}
+spec:
+  selector:
+    matchLabels:
+    {{- include "telegraf.labels" . | nindent 6 }}
+  namespaceSelector:
+    matchNames:
+    - {{ .Release.Namespace }}
+  endpoints:
+    - port: prometheus-client
+      path: /metrics
+{{- end }}
--- a/charts/telegraf/values.yaml
+++ b/charts/telegraf/values.yaml
@@ -109,6 +109,15 @@ serviceAccount:
   name:
   # Annotations for the ServiceAccount
   annotations: {}
+
+serviceMonitor:
+  # Specifies whether a ServiceMonitor should be created
+  create: false
+  # The name of the ServiceMonitor to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name:
+  # Annotations for the ServiceMonitor
+  annotations: {}
 ## Exposed telegraf configuration
 ## For full list of possible values see `/docs/all-config-values.yaml` and `/docs/all-config-values.toml`
 ## ref: https://docs.influxdata.com/telegraf/v1.1/administration/configuration/
@@ -141,6 +150,8 @@ config:
         urls:
           - "http://influxdb.monitoring.svc:8086"
         database: "telegraf"
+    - prometheus_client:
+        listen: "8080"
   inputs:
     - statsd:
         service_address: ":8125"
--- a/charts/telegraf/templates/_helpers.tpl
+++ b/charts/telegraf/templates/_helpers.tpl
@@ -707,6 +707,17 @@ Create the name of the service account to use
 {{- end -}}
 
 {{/*
+Create the name of the service monitor to use
+*/}}
+{{- define "telegraf.serviceMonitorName" -}}
+{{- if .Values.serviceMonitor.create -}}
+    {{ default (include "telegraf.fullname" .) .Values.serviceMonitor.name }}
+{{- else -}}
+    {{ default "default" .Values.serviceMonitor.name }}
+{{- end -}}
+{{- end -}}
+
+{{/*
 Get health configuration
 */}}
 {{- define "telegraf.health" -}}

### CODE REVIEW

File: charts/telegraf/templates/servicemonitor.yaml [hunk 1]
```diff
@@ -0,0 +1,22 @@
+{{- if .Values.serviceMonitor.create }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ template "telegraf.serviceMonitorName" . }}
+  labels:
+    {{- include "telegraf.labels" . | nindent 4 }}
```
alespour, line 7: I feel like it would be handy to allow users specify additional custom labels as well, for more relaxed monitor referencing from Prometheus, what do you think? I noticed Grafana template has such option.

```suggestion
    {{- include "telegraf.labels" . | nindent 4 }}
    {{- if .Values.serviceMonitor.labels }}
      {{- toYaml .Values.serviceMonitor.labels | nindent 4 }}
    {{- end }}    
```

alespour (changes requested):
Could you please bump chart version.

charts/telegraf/templates/servicemonitor.yaml
Efrat19: done, let me know when its ok

alespour (dismissed):
LGTM

alespour (dismissed):

Efrat19:
Hey is there anything else still needed to be done here? @tomklapka @alespour

charts/telegraf/values.yaml
```diff
@@ -109,6 +109,17 @@ serviceAccount:
   name:
   # Annotations for the ServiceAccount
   annotations: {}
+
+serviceMonitor:
```
tomklapka, line 113: Should be mentioned here what specific purpose is this intended for. E.g. Use specific/default outputs.prometheus_client config to match this sample settings etc.

charts/telegraf/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,25 @@
+{{- if .Values.serviceMonitor.create }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
```
tomklapka, line 4: Be aware of situation that Prometheus resources can be installed in different namespace than Telegraf: By default and before the version v0.19.0, ServiceMonitors must be installed in the same namespace as the Prometheus resource. With the Prometheus Operator v0.19.0 and above, ServiceMonitors can be selected outside the Prometheus namespace via the serviceMonitorNamespaceSelector field of the Prometheus resource. The discovered targets may come from any namespace. This allows cross-namespace monitoring use cases, for example, for meta-monitoring.

tomklapka (changes requested):

charts/telegraf/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,25 @@
...
+  {{- end }}
+spec:
+  selector:
+    matchLabels:
+    {{- include "telegraf.labels" . | nindent 6 }}
+  namespaceSelector:
+    matchNames:
+    - {{ .Release.Namespace }}
+  endpoints:
+    - port: prometheus-client
```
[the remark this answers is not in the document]
Efrat19: Done

charts/telegraf/values.yaml
Efrat19: Done

charts/telegraf/templates/servicemonitor.yaml
Efrat19: Done

Efrat19:
> Checks failed with Error: ServiceMonitor.monitoring.coreos.com "telegraf-1xv72y61rr" is invalid: spec.endpoints.port: Invalid value: "integer": spec.endpoints.port in body must be of type string: "integer". Please fix it.

fixed

charts/telegraf/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,33 @@
...
+    {{- include "telegraf.labels" . | nindent 6 }}
+  namespaceSelector:
+    matchNames:
+    - {{ .Release.Namespace }}
+  endpoints:
+  {{- range $objectKey, $objectValue := .Values.config.outputs }}
+  {{- range $key, $value := . -}}
+    {{- $tp := typeOf $value -}}
+    {{- if eq $key "prometheus_client" }}
+    - port: {{ trimPrefix ":" $value.listen }}
```
zffocussss, line 28: {{ (trimPrefix ":" $value.listen) | quote }} can correct the type issues.it is string instead of number

Issue #544, opened 2023-03-10
[telegraf] Cannot use `starlark` script as it doesn't accept multipleline string under processors.starlark.source


hungran:
Gotta issue when trying to implement [this instruction by](https://www.influxdata.com/blog/prometheus-remote-write-support-with-influxdb-2-0/)
```
[[processors.starlark]]
  ## The Starlark source can be set as a string in this configuration file, or
  ## by referencing a file containing the script.  Only one source or script
  ## should be set at once.

  ## Source of the Starlark script.
  source = '''
def apply(metric):
   if metric.name == "prometheus_remote_write":
        for k, v in metric.fields.items():
            metric.name = k
            metric.fields["value"] = v
            metric.fields.pop(k)
   return metric
'''
```
The issue I was thought it came from the template we had to wrap `processor` helm template function
Should we go with this like we did for `telegraf-ds` https://github.com/influxdata/helm-charts/pull/239
WIP

PR #545, opened 2023-03-10
fix: Add option simple to override config TOML file

hungran:
This intend to fix issue: https://github.com/influxdata/helm-charts/issues/544
Which allow to add `starlark` script under `processor`

related: https://github.com/influxdata/helm-charts/pull/239


hungran:
@powersj  I'm new here so if you have time, please have a look

PR #546, opened 2023-03-12
fix: add starlark_processor template

hungran:
This is another way to fix this issue https://github.com/influxdata/helm-charts/issues/544

### DIFF
--- a/charts/telegraf/Chart.yaml
+++ b/charts/telegraf/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: telegraf
-version: 1.8.26
+version: 1.8.27
 appVersion: 1.25.2
 deprecated: false
 description: Telegraf is an agent written in Go for collecting, processing, aggregating, and writing metrics.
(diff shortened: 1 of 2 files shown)


hungran:
@powersj can you have a look?

hungran:
could have this if we need another script for starlark

```helm
{{/*
Helper function to create a starlark processor configuration block
*/}}
{{- define "telegraf.starlark_processor" -}}
[[processors.starlark]]
{{- range $key, $value := . -}}
{{- if $value -}}
{{- $tp := typeOf $value -}}
{{- if eq $tp "string" -}}
{{ $key | nindent 2 }} = {{ $value | nindent 2 }}
{{- end }}
{{- if eq $tp "float64" }}
{{ $key }} = {{ $value | int64 }}
{{- end }}
{{- if eq $tp "int" }}
{{ $key }} = {{ $value | int64 }}
{{- end }}
{{- if eq $tp "bool" }}
{{ $key }} = {{ $value }}
{{- end }}
{{- end }}
{{- end }}
{{- end -}}
```

PR #556, opened 2023-04-27
refactor: Deduplicate env var

koooge:
Hi there, telegraf-ds's daemonset has the duplicated env var HOSTIP.

### DIFF
--- a/charts/telegraf-ds/values.yaml
+++ b/charts/telegraf-ds/values.yaml
@@ -33,13 +33,6 @@ env:
     valueFrom:
       fieldRef:
         fieldPath: spec.nodeName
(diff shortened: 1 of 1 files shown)


Bozz95:
I think this is the right choice to remove this env var from the values. 
Leaving the HOSTIP declaration only in the template file will assure this helm installation will work regardless of the resolvable host ip within the containers.

Bozz95 (approved):
I think this is the right choice to remove this env var "HOSTIP" from the values.
Leaving the HOSTIP declaration only in the template file will enure that this helm installation will work regardless of the resolvable host ip within the containers.

PR #536, opened 2023-01-18
[influxdb2] Bump to 2.7.1

centromere:

PR #563, opened 2023-06-09
influxdb2/chore: update version to v2.7.1

koooge:
Hi there, I don't know if influxdb2 is maintained, but could you update the version to the latest?

- influxdb2/chore: update version to v2.7.1


koooge:
Duped #536

PR #575, opened 2023-07-05
fix: use explicit path to container registry

jcpunk:
This helps avoid a bad actor changing your default registries to force fetching unexpected images.

My auditors prefer explicit container paths so that there is no mystery where the objects are fetch from.

### DIFF
--- a/charts/telegraf/Chart.yaml
+++ b/charts/telegraf/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: telegraf
-version: 1.8.30
+version: 1.8.31
 appVersion: 1.27.3
 deprecated: false
 description: Telegraf is an agent written in Go for collecting, processing, aggregating, and writing metrics.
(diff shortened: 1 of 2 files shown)


charts/telegraf/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: telegraf
-version: 1.8.28
+version: 1.8.29
 appVersion: 1.26.3
```
Hipska, line 4: Might as well use latest version of telegraf?
```suggestion
appVersion: 1.27.3
```

Hipska (approved):

charts/telegraf/Chart.yaml
jcpunk: Looks like somebody else beat me to the version update.

Hipska (approved):

powersj (approved):

PR #578, opened 2023-07-13
feat: prometheus-operator hooks for telegraf

jcpunk:
This adds in a simple hook for setting up a ServiceMonitor for use with Prometheus-Operator.

Configuring telegraf to expose the prometheus bits is still left as configuration for the user.


jcpunk:
https://github.com/influxdata/helm-charts/pull/510 might be a better approach...

PR #581, opened 2023-07-14
Add release notes to GH releases

morganchristiansson:
I noticed [grafana/helm-charts](https://github.com/grafana/helm-charts) repo has started to include changelog om releases and would really like to see the same in this repo!

Test release with changelog in my forked repo: https://github.com/morganchristiansson/influxdata-charts/releases/tag/telegraf-1.8.34

Changes copied from: https://github.com/grafana/helm-charts/commit/98f267b2079898e6bd23dd7b677caf98d150f6e8

Also upgraded helm version - copied from this commit https://github.com/grafana/helm-charts/commit/8d26baab28c3a85192ba42431848491ef1ba2d30


morganchristiansson:
helm 3.12.2 is out can maybe upgrade to latest. https://github.com/helm/helm/releases

I don't think helm upgraded is required tho, as you prefer.

powersj (commented):
No opposed, but can the `cr.yaml` file live under the .github/workflows directory as well?

morganchristiansson:
Isn't .github/workflows reserved for github workflows? doesn't seem like the right place.

There is already a [ct.yaml](https://github.com/influxdata/helm-charts/blob/master/ct.yaml) (used by [chart-testing](https://github.com/helm/chart-testing)) in root so this would go next to it.

Actually seems like cr.yaml in current directory is [loaded by default](https://github.com/helm/chart-releaser#config-file) so `with: config: cr.yaml` could be omitted.

powersj:
> Isn't .github/workflows reserved for github workflows? doesn't seem like the right place.

This file is referenced by the github workflow, hence why I thought it should be in there. And avoids adding a file to the root of the project. This is not something I would expect someone to run by hand right?

morganchristiansson:
chart-releaser runs great locally from CLI but you probably wouldn't except for debug/setup.

I'm not sure adding new .yaml files to .github/workflows will interfere with GitHub Actions? I'm happy to make the change if wanted tho...

PR #583, opened 2023-07-25
Added configmap and deployment annotations capabilities

Amir-Wollman:
Added ability to specify annotations in deployment and configmap.
The motivation to do so is to allow reloader to match & search resources valid for reloading, and I suspect there are further use-cases.

Did not add to service resource as it is already enabled using service.annotaions, as well as PDB, Role and RoleBinding resources for which it isn't as relevant.

### DIFF
--- a/charts/telegraf/templates/configmap.yaml
+++ b/charts/telegraf/templates/configmap.yaml
@@ -2,6 +2,10 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ include "telegraf.fullname" . }}
+  {{- if .Values.configmapAnnotations }}
+  annotations:
+  {{ toYaml .Values.configmapAnnotations | indent 4 }}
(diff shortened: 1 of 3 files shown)


SuperCoolAlan:
> Added ability to specify annotations in deployment and configmap. The motivation to do so is to allow reloader to match & search resources valid for reloading, and I suspect there are further use-cases.

@Amir-Wollman when you mention the "reloader" here, which are you referring to?

Amir-Wollman:
> > Added ability to specify annotations in deployment and configmap. The motivation to do so is to allow reloader to match & search resources valid for reloading, and I suspect there are further use-cases.
> 
> @Amir-Wollman when you mention the "reloader" here, which are you referring to?

Sorry for the delayed response - didn't notice this message.
By reloader, I mean the stakater reloader, but this is probably useful for more implementations.

PR #599, opened 2023-09-04
[Influxdb] Support containerSecurityContext

PaulFarver:
Support securityContext on container level in influxdb helm chart


AlexanderThaller:
Would need this as well

PR #634, opened 2024-04-15
feat: [influxdb3-clustered] add the initial chart

jenniferplusplus:
Adds the InfluxDB 3.0 Clustered chart, and some basic usage documentation


charts/influxdb3-clustered/values.yaml
```diff
@@ -0,0 +1,222 @@
...
+
+nameOverride: ""
+namespaceOverride: ""
+fullnameOverride: ""
+kubitSingleNamespace: false
+operatorImageVersion: "v0.0.14"
+operatorImageOverride: ""
+
+# set to true if you have previously installed the kubit operator from another source
+skipOperator: false
```
jdockerty, line 16: I imagine we going to place your internal write up on helm/kubit adoption to the official docs site?

Having a link to that once it is available will be really useful I think here :+1:

charts/influxdb3-clustered/values.yaml
```diff
@@ -0,0 +1,222 @@
...
+  # If you are using cert-manager, enter a name for the Secret it should create.
+  # tlsSecretName: ""
+  http:
+    # Usually you have only one ingress controller installed in a given cluster.
+    # In case you have more than one, you have to specify the "class name" of the ingress controller you want to use
+    className: nginx
+  grpc:
+    # Usually you have only one ingress controller installed in a given cluster.
+    # In case you have more than one, you have to specify the "class name" of the ingress controller you want to use
+    className: nginx
```
jdockerty, line 126: I wonder whether the default here should be blank as opposed to `nginx`?

I do think this is fine, since the default of the [nginx helm chart also uses the `nginx` class](https://github.com/kubernetes/ingress-nginx/blob/b9e0c281d060ff3030daf300790a04e0fdb89a2e/charts/ingress-nginx/values.yaml#L119-L123) anyway. Plus, `ingress` here is optional.

This is mainly something which came to mind after looking over it again, not a deal breaker I don't think :smile:

jdockerty (dismissed):
Looks good to me, few comments from looking over this again.

I think it'll be useful to get some more eyes on and testing it out of this repo (from this branch) as well just to be sure we're 1:1.

charts/influxdb3-clustered/values.yaml
jenniferplusplus: Yeah, I think that's why I chose nginx as the default. But we could leave them empty, since it's optional.

jdockerty (approved):

waynr (approved):
:shipit:

PR #558, opened 2023-05-04
feat: Add imagePullSecrets to telegraf-ds

koooge:
Hi there,  Add imagePullSecrets to telegraf-ds 

- refs: https://github.com/influxdata/helm-charts/commit/65823bc069c1f5c96a90832c0687efccf7ca679d
- refs: #482

PR #640, opened 2024-05-01
[telegraf-ds] add imagePullSecrets support

jmickey:
Add support for `imagePullSecrets` in `telegraf-ds`. With the DockerHub pull rate limiting it's important that users are able to provide auth creds for DockerHub.

There is a previous MR: https://github.com/influxdata/helm-charts/pull/558, but it is quite old and out of date.


jenniferplusplus (approved):
Looks good, thanks

PR #644, opened 2024-05-13
Make replicaCount optional when set value to null

Jaydee94:
This pr makes the replica count optional in the telegraf Deployment object.

This is useful when you use a HorizontalPodAutoscaler that manages the replica count for you. Especially when you deploy this helm chart with ArgoCD and the auto sync functionality (https://argo-cd.readthedocs.io/en/stable/user-guide/best_practices/#leaving-room-for-imperativeness).

I've successfully tested it locally on my machine with `helm template . --set replicaCount=null`


Jaydee94:
Hi, is there a chance that this is being accepted? 😊

PR #646, opened 2024-05-16
Add possibility to set priorityClassName on InfluxDB Pod via Helmchart

moandersson:
https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass


moandersson:
Is this project still active? What is the process of getting this reviewed?

PR #662, opened 2024-08-12
feat: support define telegraf container ports

neilkuan:
Support define telegraf container ports for success as  [victoria metrics agent](https://github.com/VictoriaMetrics/helm-charts/tree/4040c7e6f9a134972535fb1fdf9c973a2dfb53be/charts/victoria-metrics-agent) scrape targets.
- default drop rule in [kubernetes-service-endpoints-job](https://github.com/VictoriaMetrics/helm-charts/blob/4040c7e6f9a134972535fb1fdf9c973a2dfb53be/charts/victoria-metrics-agent/values.yaml#L422-L426)

### DIFF
--- a/charts/telegraf/values.yaml
+++ b/charts/telegraf/values.yaml
@@ -181,7 +181,7 @@ pdb:
   minAvailable: 1
   # maxUnavailable: 1
 
-containerPorts:
-  - name: http
-    containerPort: 9273
-    protocol: TCP
+# containerPorts:
+# - name: http
+#   containerPort: 9273
+#   protocol: TCP
--- a/charts/telegraf/values.yaml
+++ b/charts/telegraf/values.yaml
@@ -180,3 +180,8 @@ pdb:
   create: true
   minAvailable: 1
   # maxUnavailable: 1
+
+containerPorts:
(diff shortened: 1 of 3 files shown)


neilkuan:
Hi @powersj , could you take a look at?! Thanks.

charts/telegraf/values.yaml
```diff
@@ -180,3 +180,8 @@ pdb:
   create: true
   minAvailable: 1
   # maxUnavailable: 1
+
+containerPorts:
+  - name: http
+    containerPort: 9273
+    protocol: TCP
```
powersj, line 187: @neilkuan thanks for the PR! I'm happy to take this I think, although I'm a little surprised no one has ever requested this before.

In either case, shouldn't this section be commented out? Otherwise a user with the default config would always have port 9273 exposed?

powersj (approved):

charts/telegraf/values.yaml
neilkuan: Good catch. 
I will comment out it. Thanks for Review.

neilkuan:
Thanks for review @powersj. 😎

PR #671, opened 2024-08-28
fix(edge): Fixes enterprise helm chart to allow config of livenessProbe

devanbenz:
- Per our helm chart documentation consumers should be able to configure the livenessProbe
- This adds functionality to let users actually do that


charts/influxdb-enterprise/values.yaml
```diff
@@ -23,6 +23,14 @@ serviceAccount:
   name: ''
   annotations: {}
 
+## Customize liveness, readiness and startup probes
+## ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
+livenessProbe: {}
+  # path: "/ping"
+  # initialDelaySeconds: 30
+  # timeoutSeconds: 5
+  # scheme: HTTP
```
jdockerty, line 32: Since this `scheme` isn't consumed I think this might result in an YAML error for the templating engine through `helm`.

I've not tried this myself, but at a quick glance seeing that `.Values.livenessProbe.scheme` isn't used in the `scheme` field might be problematic. I suspect the default can be `HTTP` in a similar manner to the `/ping` path

charts/influxdb-enterprise/templates/data-statefulset.yaml
jdockerty, line 1: Does the same also need to be placed onto the meta nodes too?

Perhaps that wasn't the original ask, but a slight rejig to allowing something like:

```yaml
# values.yaml
data:
  livenessProbe:
     ...

meta:
  livenessProbe:
    ...
```

might be quite worthwile

---

For reference, as the `livenessProbe` field exists in both files:

```
rg "livenessProbe" templates -l
templates/meta-statefulset.yaml
templates/data-statefulset.yaml
```

jdockerty (commented):
LGTM, small things to check before :heavy_check_mark:

charts/influxdb-enterprise/templates/data-statefulset.yaml
```diff
@@ -117,10 +117,11 @@ spec:
               protocol: UDP
           livenessProbe:
             httpGet:
-              path: /ping
+              path: {{ .Values.livenessProbe.path | default "/ping" }}
```
jenniferplusplus, line 120: This is surprising to me. What other path would be more useful as a liveness probe? And what are the implications of this path being different than the path on the readiness probe?

charts/influxdb-enterprise/values.yaml
jenniferplusplus: Same with timeoutSeconds

charts/influxdb-enterprise/templates/data-statefulset.yaml
devanbenz: This was more so for additional functionality. I can remove it!

charts/influxdb-enterprise/templates/data-statefulset.yaml
devanbenz: I'll go ahead and add to the meta nodes too

devanbenz: @jdockerty should I be splitting up the livenessProbe or maybe just have a single value in the helm chart? If customers indicate they want it split we can do so but as of right now I think the customer is just wanting to adjust the data nodes.

jdockerty: > customer is just wanting to adjust the data nodes

In this case I think going to satisfy what a customer wants is fine, as it is a quick win, although it isn't my call too unfortunately. Might be worth checking with Geoffrey on that one :smile:

devanbenz:
@jdockerty @jenniferplusplus I've modified this pr to only adjust per the requirement from our customer 🫡

charts/influxdb-enterprise/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-version: 0.1.22
+version: 0.1.24
```
jdockerty, line 2: Does this need a bump for 2 minor versions, I think simply `0.1.23` here is sufficient.

devanbenz: Check was failing for 1.23 😭 I'll try again but I was just listening to CI 😆

devanbenz: Cool it worked -- not sure why it was having issues previously. Maybe it was something not related 😅

jenniferplusplus (approved):

PR #673, opened 2024-09-04
fix(edge): need to set end block for http

devanbenz:
- typo when creating the initial PR
- Need to set up the end before the liveness probe or else it gets caught up in the https enabled false and you cannot set it


jdockerty (approved):
A simple fix :rocket:

PR #685, opened 2024-10-16
chore: Bump github action versions

srebhan:
This is an attempt to fix the recent workflow errors complaining about python v3.7 not being available for Ubuntu 24.04. While being at it, this PR bumps all actions to their latest available version.

### DIFF
--- a/.github/workflows/helm-charts-test.yaml
+++ b/.github/workflows/helm-charts-test.yaml
@@ -8,42 +8,43 @@ jobs:
     strategy:
       matrix:
         node_image_version:
-          - v1.21.1 # default used by helm/kind-action@v1.2.0 (kubernetes-sigs/kind@0.11.1)
+          - v1.21.1
           - v1.22.4
     name: lint-test (k8s ${{ matrix.node_image_version }})
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/helm-charts-test.yaml
```diff
@@ -8,42 +8,43 @@ jobs:
...
-        uses: azure/setup-helm@v1
+        uses: azure/setup-helm@v4
         with:
           version: v3.6.3
 
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@v5
         with:
-          python-version: 3.7
+          python-version: '3.x'
```
jdstrand, line 28: Ah interesting. For posterity, @srebhan and I discussed this elsewhere and with the current `ubuntu-latest` (24.04), the earliest version is 3.8, which is fine.

jdstrand (approved):
LGTM, thanks!

.github/workflows/helm-charts-test.yaml
srebhan: I checked and python is used for `ct lint` as stated in the [helm/chart-testing-action readme](https://github.com/helm/chart-testing-action?tab=readme-ov-file#example-workflow). So I think there is very little that could go wrong there. (Famous last words... ;-))

PR #699, opened 2025-01-16
Revert "chore: helm chart clustered update"

jenniferplusplus:
Reverts influxdata/helm-charts#695

This release contains a defect that can, in some cases, cause data loss. It should not be used.

### DIFF
--- a/charts/influxdb3-clustered/Chart.yaml
+++ b/charts/influxdb3-clustered/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 type: application
 
-version: 0.1.9
-appVersion: "20241217-1494922"
+version: 0.1.8
+appVersion: "20240925-1257864"
 name: influxdb3-clustered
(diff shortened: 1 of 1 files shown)


charts/influxdb3-clustered/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 type: application
 
-version: 0.1.9
-appVersion: "20241217-1494922"
+version: 0.1.8
+appVersion: "20240925-1257864"
```
jenniferplusplus, line 5:
```suggestion
version: 0.1.10
appVersion: "20240925-1257864"
```

wdoconnell (approved):

jenniferplusplus:
We determined that the defect does not affect Clustered deployments, and so this release is still safe to use.

The bug is triggered by a certain partitioning scheme that predates the release of the Clustered product, and has never been available in Clustered.

PR #718, opened 2025-06-16
feat: add the ability to pause kubit operator in appinstance template

Abhi94N:
Hi, I wanted to add the ability to pause the kubit operator from the helm chart.


jenniferplusplus (approved):
nice, thank you!

PR #729, opened 2025-07-27
feat: add option for namespace selection for telegraf-operator webhook

SHA65536:
I added an option for the webhook to only match certain namespaces instead of all of them
In GKE they complain when you have a webhook that can intercept requests for system namespaces (kube-system, kube-node-lease) so I needed a way to adhere to their [guidelines](https://cloud.google.com/kubernetes-engine/docs/how-to/optimize-webhooks#unsafe-webhooks)
I tested this on our company infra, seems fine
Thanks!


SHA65536:
@skartikey @eatondustin1  can you take a look?
Sorry for the ping, saw you reviewed stuff recently

PR #742, opened 2025-09-25
feat: [telegraf] add support for initContainers

paulojmdias:
This PR adds support for defining initContainers in the Telegraf Helm chart. It allows users to run one or more initialization containers before Telegraf starts, enabling common use cases such as waiting for dependencies, preparing configuration files, or setting permissions.

Example:

```yaml
initContainers:
  - name: init-myservice
    image: busybox:1.36
    command: ["sh", "-c", "echo 'Init container ran successfully'"]
```

### DIFF
--- a/.github/workflows/helm-charts-test.yaml
+++ b/.github/workflows/helm-charts-test.yaml
@@ -31,6 +31,8 @@ jobs:
       - name: Install chart-testing tools
         id: lint
         uses: helm/chart-testing-action@v2
+        with:
+          yamale_version: '6.0.0'
 
       - name: Run chart-testing (list-changed)
         id: list-changed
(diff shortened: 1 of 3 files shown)


charts/telegraf/values.yaml
```diff
@@ -16,6 +16,9 @@ args: []
 # be added to the environment (must be manually created)
 # This can be useful for auth tokens, etc.
 
+## Additional initContainers to initialize the pod
```
skartikey: How about we enhance the comment, something like:
```suggestion
## Init containers to run before Telegraf starts
## Useful for waiting on dependencies, preparing configs, or setting permissions
## Example:
# initContainers:
#   - name: wait-for-service
#     image: busybox:1.36
#     command: ['sh', '-c', 'until nc -z myservice 8086; do sleep 1; done']
```

skartikey (approved):
@paulojmdias Thanks for the contribution!

PR #743, opened 2025-10-01
fix: remove default docker_endpoint for K8s 1.24+ compatibility

heliapb:
Screenshot 2025-10-01 at 19 20 29

My team is facing an issue due to defined value that is blocking the k8s upgrade and is causing an error. Ideally this should not be defined on the main values, but instead be defined if needed to the proper value.

### DIFF
--- a/charts/telegraf-ds/Chart.yaml
+++ b/charts/telegraf-ds/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: telegraf-ds
-version: 1.1.42
+version: 1.1.43
 appVersion: 1.36.1
 deprecated: false
 description: Telegraf is an agent written in Go for collecting, processing, aggregating, and writing metrics.
--- a/charts/telegraf-ds/README.md
+++ b/charts/telegraf-ds/README.md
@@ -80,3 +80,12 @@ This chart deploys the following by default:
 The following plugin can be enable through the configuration:
 
 - [`internal`](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/internal)
+
+## Docker metrics
+
+To enable Docker monitoring (if Docker runtime is available):
+
+```yaml
+config:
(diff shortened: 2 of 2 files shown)


heliapb:
Hi @srebhan @bnpfeife @skartikey when you have some time pls review, thanks.

skartikey:
@heliapb Thanks for the PR!

The change is correct, both `configmap.yaml` and `daemonset.yaml` handle an empty `docker_endpoint` with conditional logic. The existing test `ci/docker-no-endpoint-values.yaml` already covers this case.

A few general comments:

Please increment the version in `charts/telegraf-ds/Chart.yaml` to trigger a release.

**Breaking change:**
  This update affects users relying on Docker metrics. Please update the chart README to document why the default changed and how to explicitly enable Docker monitoring if needed (for users still on Docker runtime):

  ```yaml
  # To enable Docker monitoring (if Docker runtime is available):
  config:
    docker_endpoint: "unix:///var/run/docker.sock"
  ```

Commit message:
Current: `fix: depreacated ci` (has a typo and doesn’t describe the actual fix)
Suggested: `fix: remove default docker_endpoint for K8s 1.24+ compatibility`

Optional:

Consider adding a comment in `values.yaml` explaining why `docker_endpoint` is empty:

  ```yaml
  # Docker socket endpoint for Docker input plugin
  # Empty by default as Docker (dockershim) was removed in K8s 1.24+
  # Set to "unix:///var/run/docker.sock" only if using Docker runtime
  docker_endpoint: ""
  ```

skartikey:
@bnpfeife, please review before we proceed with the merge.

bnpfeife (approved):

bnpfeife:
My apologies, I didn't see this until just now. I rendered the helm chart locally (using this branch), and it worked as expected. So, it looks good to me.

PR #757, opened 2025-12-24
feat(influxdb2): Add Gateway API HTTPRoute support

younsl:
Add HTTPRoute resource as an alternative to Ingress for exposing InfluxDB via Kubernetes Gateway API.

- Add httproute.yaml template
- Add httpRoute configuration in values.yaml
- Update README.md with HTTPRoute usage example


younsl:
@alespour PTAL. Once this lands, users can set up HTTPRoute just through values.yaml without having to write and maintain a separate `httproute.yaml` in their `templates/` directory.

PR #769, opened 2026-03-05
fix(influxdb3-ent): add gRPC/Flight Ingress route

alespour:
This PR 

Adds gRPC/Flight Ingress routing to InfluxDB3 Enterprise chart. It is necessary for querying with v3 client libraries. Keeps a single default ingress host for path-based write/query/Flight routing; an optional `ingress.flight.host` override is available for ingress-controller environments that require host-level separation of gRPC and HTTP traffic.

### DIFF
--- a/charts/influxdb3-enterprise/README.md
+++ b/charts/influxdb3-enterprise/README.md
@@ -91,15 +91,13 @@ At minimum, you must configure:
      # For commercial licenses use `file: /path/to/license` or `existingSecret`
    ```
 
-3. **Ingress Hosts**:
+3. **Ingress Host**:
    ```yaml
    ingress:
-     write:
-       hosts:
-         - host: writes.your-domain.com
-     query:
-       hosts:
-         - host: query.your-domain.com
+     host: "influxdb.example.com"
+     # Optional Flight host override (defaults to ingress.host)
+     flight:
+       host: ""
    ```
 
 ## Configuration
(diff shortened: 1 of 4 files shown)


charts/influxdb3-enterprise/templates/ingress-flight.yaml
```diff
@@ -0,0 +1,58 @@
...
+            # Keep ImplementationSpecific for dotted gRPC service paths.
+            pathType: ImplementationSpecific
+            backend:
+              service:
+                name: {{ include "influxdb3-enterprise.fullname" . }}-querier
+                port:
+                  number: {{ .Values.querier.service.port }}
+          - path: /arrow.flight.protocol.sql.FlightSqlService
+            # Keep ImplementationSpecific for dotted gRPC service paths.
+            pathType: ImplementationSpecific
```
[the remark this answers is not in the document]
alespour: Using `pathType: ImplementationSpecific` is intentional for Flight gRPC routes.

With the default chart ingress class (`ingress-nginx`), using `Prefix` for dotted gRPC service paths (`/arrow.flight.protocol.FlightService`, `/arrow.flight.protocol.sql.FlightSqlService`) triggers controller warnings; validated directly in-cluster. The path is not a regex; the dots are literal characters in the gRPC service name. However, `ingress-nginx` validates `Prefix` paths more strictly and warns/rejects dotted paths. So `ImplementationSpecific` is the correct `pathType` for this literal dotted gRPC route.

On precedence: in `ingress-nginx`, path matching is longest-path-first, so `/arrow.flight.protocol...` is more specific than the query ingress / path and should take precedence on the same host. Precedence inversion was not observed in testing. 

`ingress.flight.host` / `ingress.flight.tls` are kept as compatibility escape hatches for environments where a specific ingress/controller requires host-level separation of gRPC and HTTP traffic.

alespour:
Tested with JS/Go/Java clients.

charts/influxdb3-enterprise/templates/ingress-flight.yaml
```diff
@@ -0,0 +1,58 @@
...
+  {{- if .Values.ingress.flight.tls }}
+  tls:
+    {{- range .Values.ingress.flight.tls }}
+    - hosts:
+        {{- range .hosts }}
+        - {{ . | quote }}
+        {{- end }}
+      secretName: {{ .secretName }}
+    {{- end }}
+  {{- else if .Values.ingress.tls }}
```
bednar: When `ingress.flight.host` differs from `ingress.host`, this fallback copies the global TLS block unchanged. That renders a Flight ingress with `rules.host` set to the Flight host, but `spec.tls[].hosts` still pointing at the normal ingress host, so TLS will not match the documented host override unless users also discover and set `ingress.flight.tls`. Please either only reuse `ingress.tls` when the effective Flight host matches `ingress.host`, or require/document an explicit `ingress.flight.tls` whenever `ingress.flight.host` is set.

charts/influxdb3-enterprise/README.md
```diff
@@ -594,8 +604,17 @@ logs:
 | `ingress.className` | Ingress class | `nginx` |
 | `ingress.port` | Host port referenced in NOTES | `8181` |
 | `ingress.tls` | TLS host/secret list | `[]` |
-
-Ingress paths are fixed in the templates: write ingress exposes `/api/v3/write_lp`, `/api/v2/write`, `/write`; query ingress exposes `/api/v3/query`, `/query`, and `/`.
+| `ingress.write.annotations` | Write ingress annotations | `proxy-body-size/read-timeout` |
+| `ingress.query.annotations` | Query ingress annotations | `proxy-read-timeout` |
+| `ingress.flight.host` | Optional host override for Flight ingress | `""` (uses `ingress.host`) |
+| `ingress.flight.tls` | Optional TLS override for Flight ingress | `[]` (uses `ingress.tls`) |
```
bednar: This default is technically true, but it is unsafe for the documented separate-host use case: setting `ingress.flight.host` while inheriting `ingress.tls` leaves the TLS hosts on `ingress.host`. If the template continues to inherit global TLS, please call out here that a distinct Flight host must be paired with a matching `ingress.flight.tls` entry.

bednar (changes requested):
Requesting changes for the Flight host/TLS override edge case. The rendered manifests can currently produce a Flight ingress whose rule host and TLS hosts do not match when `ingress.flight.host` is set but `ingress.flight.tls` is omitted.

Also, the PR body still says this bumps the app version to latest `3.9.1`, but the rebased diff now keeps `appVersion: 3.9.2` and only bumps the chart version to `0.5.0`. Please update the description before merge.

charts/influxdb3-enterprise/templates/ingress-flight.yaml
alespour: fixed in https://github.com/influxdata/helm-charts/pull/769/commits/c3cc39c6895048b5433da272d3f1ccd70a8effb1

charts/influxdb3-enterprise/README.md
alespour: addressed in https://github.com/influxdata/helm-charts/pull/769/commits/c3cc39c6895048b5433da272d3f1ccd70a8effb1

alespour:
> Also, the PR body still says this bumps the app version to latest `3.9.1`, but the rebased diff now keeps `appVersion: 3.9.2` and only bumps the chart version to `0.5.0`. Please update the description before merge.

This PR is not the most recent so app version drifted already. Removed explicit mention of any version bump.

bednar (approved):
Reviewed the update. The previous Flight host/TLS mismatch is fixed, the related documentation now calls out the separate-host TLS requirement, and the rendered Helm manifests/checks look good.

PR #776, opened 2026-04-01
fix(influxdb-ent): TLS secret permissions

alespour:
This PR
* fixes TLS key-permission startup failures introduced by stricter certificate/key permission checks in v1.12.2+ by mounting TLS secrets with secure mode 0600 for meta and data pods
* adds optional `insecureCertificate` values (default `false`) that maps to `https-insecure-certificate`  v1.12.3+ configuration option, just for compatibility edge cases.


rkinginflux (approved):
LGTM

bednar (approved):

PR #777, opened 2026-04-02
fix(telegraf-ds|telegraf): add tpl support for podAnnotations

paulojmdias:
Allow Helm template expressions in `podAnnotations` by wrapping the existing `toYaml` call with `tpl()` in both `telegraf` and `telegraf-ds`.

This is backward compatible. Static annotation values are unaffected.


skartikey (approved):
@paulojmdias Thanks!

PR #807, opened 2026-06-26
fix(influxdb3-ent): remove unnecessary ingester WAL PVC

### LINKED ISSUE #806: [influxdb3-enterprise] drop unnecessary WAL PVC
The chart creates a dedicated WAL `PersistentVolumeClaim` for ingesters, but InfluxDB 3 Enterprise stores the WAL in the configured object store, not on a local volume. The PVC is unnecessary for remote object storage (S3, Azure, GCS).

Fix: Remove the ingester WAL PVC. WAL is persisted through the object store. For `objectStorage.type=file`, mount the existing shared object-storage PVC instead.

alespour:
Closes #806 

Removes the dedicated ingester WAL `PersistentVolumeClaim`. InfluxDB 3 Enterprise persists the WAL in the configured object store, so a separate volume isn't needed. Per the [durability docs](https://docs.influxdata.com/influxdb3/enterprise/reference/internals/durability/), no separate local disk volume is required for the WAL itself since it lives in the object store.

Changes:
  - Drops the ingester WAL `volumeClaimTemplate` and `mount`.
  - For `objectStorage.type=file`, mounts the shared object-storage PVC at `dataDir` across all components (previously declared but never mounted).
  - Makes the file object-storage PVC access mode configurable (`objectStorage.file.persistence.accessMode`, default `ReadWriteMany`).
  - Deprecates `ingester.persistence.*` (kept for compatibility, unused).
  -  Documents the 0.6.x upgrade path for the immutable `StatefulSet` change, including old WAL PVC cleanup and the `objectStorage.type=file` backup caveat. Object store of `file` type is dev/test-only, not recommended for production, which simplifies upgrade (no WAL data migration).

### DIFF
--- a/charts/influxdb3-enterprise/README.md
+++ b/charts/influxdb3-enterprise/README.md
@@ -481,16 +481,22 @@ After the upgraded ingester pods are healthy, delete the old ingester WAL PVCs.
 Chart 0.6.x created these PVCs when `ingester.persistence.enabled` was true, but
 they were unused leftovers: InfluxDB 3 Enterprise writes the WAL to the
 configured object store, not to those local PVC mounts. There is one old WAL PVC
-per ingester replica.
+per ingester replica. The PVC name pattern is
+`wal-<ingester-statefulset-name>-<ordinal>`, where `<ordinal>` starts at `0`.
 
 ```bash
-kubectl get pvc -n influxdb3 -l app.kubernetes.io/component=ingester
-# Default release name examples:
+# List old WAL PVCs for the default StatefulSet name and namespace:
+kubectl get pvc -n influxdb3 | grep '^wal-influxdb3-enterprise-ingester-'
+
+# Delete one old WAL PVC per old ingester replica:
 # wal-influxdb3-enterprise-ingester-0
 # wal-influxdb3-enterprise-ingester-1
 kubectl delete pvc -n influxdb3 wal-influxdb3-enterprise-ingester-0
 kubectl delete pvc -n influxdb3 wal-influxdb3-enterprise-ingester-1
 ```
+
+Replace `influxdb3-enterprise-ingester` with your old ingester StatefulSet name
+and delete one PVC for each old ingester replica ordinal.
 
 ### Rolling Updates
 
--- a/charts/influxdb3-enterprise/README.md
+++ b/charts/influxdb3-enterprise/README.md
@@ -461,15 +461,14 @@ Chart 0.7.0 removes the ingester WAL `volumeClaimTemplates` from the
 StatefulSet. Kubernetes does not allow this field to be removed from an existing
 StatefulSet during an in-place `helm upgrade`.
 
-For existing 0.6.x releases using `objectStorage.type=file`, back up
-`objectStorage.file.dataDir` from the ingester pods before deleting them. File
-storage is intended for development/testing only, and chart 0.6.x did not mount
-the shared object-storage PVC at that path.
+For existing releases installed with chart 0.6.x, delete the ingester
+StatefulSet before upgrading. This is required for all object-store types
+because the immutable `volumeClaimTemplates` field changed.
 
-For existing releases installed with chart 0.6.x, schedule a maintenance window,
-stop writes, and allow the WAL to flush to the configured object store. Then
-delete the ingester StatefulSet before upgrading. The old WAL PVCs are retained
-by Kubernetes.
+For `objectStorage.type=file` only, chart 0.6.x stored object-store contents on
+the pod's ephemeral filesystem. File storage is intended for development/testing
+only. If you used file storage and want to preserve that dev/test data, back up
+`objectStorage.file.dataDir` from a live pod before deleting the StatefulSet.
 
 ```bash
 kubectl delete statefulset -n influxdb3 influxdb3-enterprise-ingester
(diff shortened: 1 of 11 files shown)


bednar (changes requested):
Thanks for the fix. I found a few issues that should be addressed before merging:

1. **Add an upgrade/migration path for existing installations.**
   This PR removes `volumeClaimTemplates` from the ingester StatefulSet, while chart `0.6.0` rendered it by default via `ingester.persistence.enabled=true`. Kubernetes does not allow removing `volumeClaimTemplates` from an existing StatefulSet with an in-place update, so `helm upgrade` for existing releases is likely to fail on the immutable StatefulSet spec. Please document a safe upgrade procedure, for example deleting/recreating or orphaning the ingester StatefulSet and explaining what to do with the old WAL PVCs, or use another compatibility strategy.

2. **Fix the persistence wording for `memory` and `memory-throttled`.**
   The README currently groups `memory` and `memory-throttled` with real object stores and says WAL files, snapshots, catalog data, and Parquet files are persisted through the configured object store. That is misleading for memory-backed storage because it is ephemeral. Please split or reword that paragraph so only persistent object stores are described as durable.

3. **Update stale WAL/PVC troubleshooting docs.**
   The README and TROUBLESHOOTING docs still refer to checking the ingester WAL PVC and increasing `ingester.persistence.size`, but this PR makes `ingester.persistence.*` compatibility-only and unused by the default templates. Please update or remove those sections so the docs match the new persistence model.

charts/influxdb3-enterprise/templates/ingester-statefulset.yaml
```diff
@@ -151,18 +143,4 @@ spec:
         {{ include "influxdb3-enterprise.sharedVolumes" . | nindent 8 }}
         {{ include "influxdb3-enterprise.adminTokenVolumes" . | nindent 8 }}
         {{ include "influxdb3-enterprise.permissionTokensVolumes" . | nindent 8 }}
-  {{- if .Values.ingester.persistence.enabled }}
-  volumeClaimTemplates:
```
bednar, line 155: Removing the `volumeClaimTemplates` block changes an immutable StatefulSet field for existing releases. Chart 0.6.0 rendered this by default via `ingester.persistence.enabled=true`, so an in-place `helm upgrade` can fail when Kubernetes rejects the StatefulSet update. Please add a documented migration path, for example deleting/recreating or orphaning the ingester StatefulSet and explaining what to do with the old WAL PVCs, or use another compatibility strategy.

charts/influxdb3-enterprise/README.md
```diff
@@ -239,12 +239,23 @@ objectStorage:
...
+  file:
+    dataDir: "/var/lib/influxdb3"
+    persistence:
+      accessMode: ReadWriteMany
+      size: 100Gi
 ```
 
+For `objectStorage.type=s3`, `google`, `azure`, `memory`, or `memory-throttled`,
+the chart does not create a data or WAL PVC. WAL files, snapshots, catalog data,
+and Parquet files are persisted through the configured object store.
```
bednar: This wording groups `memory` and `memory-throttled` with durable object stores and says WAL files, snapshots, catalog data, and Parquet files are persisted through the configured object store. For memory-backed storage that is misleading because the object store is ephemeral. Please split or reword this so only durable backends are described as persistent.

charts/influxdb3-enterprise/README.md
```diff
@@ -654,10 +665,11 @@ Ingress routes:
 
 | Parameter | Description | Default |
 |-----------|-------------|---------|
-| `ingester.persistence.enabled` | Enable WAL PVC | `true` |
-| `ingester.persistence.size` | WAL PVC size | `10Gi` |
+| `ingester.persistence.enabled` | Deprecated compatibility value; WAL is persisted through the configured object store | `false` |
+| `ingester.persistence.size` | Deprecated compatibility value; not used by default templates | `10Gi` |
```
bednar: Since this PR makes `ingester.persistence.*` compatibility-only and unused by the default templates, the remaining WAL/PVC troubleshooting docs should be updated too. The README still tells users to check PVC status for ingester WAL issues, and `TROUBLESHOOTING.md` still shows the old `data-influxdb3-enterprise-ingester-0` PVC plus an `ingester.persistence.size` fix. Please update or remove those stale sections so the docs match the new persistence model.

charts/influxdb3-enterprise/README.md
```diff
@@ -440,6 +455,43 @@ helm upgrade influxdb3-enterprise . \
...
+```
+
+After the upgraded ingester pods are healthy, delete the old ingester WAL PVCs.
+Chart 0.6.x created these PVCs when `ingester.persistence.enabled` was true, but
+they were unused leftovers: InfluxDB 3 Enterprise writes the WAL to the
+configured object store, not to those local PVC mounts. There is one old WAL PVC
+per ingester replica.
+
+```bash
+kubectl get pvc -n influxdb3 -l app.kubernetes.io/component=ingester
```
bednar: This selector likely does not find the old WAL PVCs created by chart 0.6.x. The old `volumeClaimTemplates` only set `metadata.name: wal`; they did not add `app.kubernetes.io/component=ingester` labels to the generated PVCs. Please replace this with a label-free lookup/deletion procedure based on the actual claim-name pattern, and call out the release name / replica count assumptions so users can reliably clean up the old PVCs.

bednar (changes requested):
One additional issue in the new upgrade documentation should be fixed before merging.

charts/influxdb3-enterprise/README.md
alespour: addressed in https://github.com/influxdata/helm-charts/pull/807/commits/769f20fc97cce02467c336bf20032acc4495cb10

bednar (approved):
Looks good. Previous requested changes have been addressed, review threads are resolved, and CI is green.

PR #812, opened 2026-07-02
fix(influxdb3-ent): avoid duplicate INFLUXDB3_UNSET_VARS

alespour:
This PR changes rendering of `INFLUXDB3_UNSET_VARS`. Renders non-processor pods `INFLUXDB3_UNSET_VARS` env var, introduced in #794, only when users do not override it through `extraEnv`. This avoids duplicate env entries while keeping user-provided low-level unset vars authoritative.

Reasoning: 

`INFLUXDB3_UNSET_VARS` is a low-level, expert knob. A user who sets it is deliberately reaching under the hood. The possible alternative "merge" approach would silently mutate the user's value (appending `INFLUXDB3_PLUGIN_DIR` to their value), which would lead to surprise someone who expects their `extraEnv` to be honored verbatim. Also, implementation would be non-trivial.
"Override-wins" avoids that surprise and keeps the code simple and consistent with the rest of the `extraEnv` model.

### DIFF
--- a/charts/influxdb3-enterprise/README.md
+++ b/charts/influxdb3-enterprise/README.md
@@ -368,11 +368,13 @@ extraEnv:
 Use component-specific `extraEnv` to target only one component. Component-specific entries override top-level `extraEnv` entries with the same `name`.
 
 ```yaml
-processingEngine:
+ingester:
   extraEnv:
-    - name: INFLUXDB3_UNSET_VARS
-      value: "INFLUXDB3_PLUGIN_DIR"
+    - name: CUSTOM_INGESTER_VAR
+      value: "custom-value"
 ```
+
+Non-processor pods set `INFLUXDB3_UNSET_VARS=INFLUXDB3_PLUGIN_DIR` by default to keep the Processing Engine disabled. If you override `INFLUXDB3_UNSET_VARS` through top-level or component `extraEnv`, include `INFLUXDB3_PLUGIN_DIR` yourself or the Processing Engine may initialize on those pods.
 
 #### Monitoring
 
--- a/charts/influxdb3-enterprise/README.md
+++ b/charts/influxdb3-enterprise/README.md
@@ -374,7 +374,9 @@ ingester:
       value: "custom-value"
 ```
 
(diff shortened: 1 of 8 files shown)


charts/influxdb3-enterprise/README.md
```diff
@@ -368,12 +368,14 @@ extraEnv:
...
-processingEngine:
+ingester:
   extraEnv:
-    - name: INFLUXDB3_UNSET_VARS
-      value: "INFLUXDB3_PLUGIN_DIR"
+    - name: CUSTOM_INGESTER_VAR
+      value: "custom-value"
 ```
 
+Non-processor pods set `INFLUXDB3_UNSET_VARS=INFLUXDB3_PLUGIN_DIR` by default to keep the Processing Engine disabled. If you override `INFLUXDB3_UNSET_VARS` through top-level or component `extraEnv`, include `INFLUXDB3_PLUGIN_DIR` yourself or the Processing Engine may initialize on those pods.
```
bednar: Please clarify the top-level `extraEnv` behavior here. A top-level `INFLUXDB3_UNSET_VARS` override is rendered into the processor StatefulSet too, so following this guidance with `processingEngine.enabled=true` and including `INFLUXDB3_PLUGIN_DIR` would also unset the plugin dir on processor pods and disable the Processing Engine there. Please either steer users toward component-specific overrides for non-processor pods, or explicitly warn that this top-level override should not be used when processor pods are enabled unless that side effect is intended.

bednar (changes requested):
The templating change looks correct in the rendered cases I checked. I am requesting one documentation clarification before approval, because the new guidance can lead users with the Processing Engine enabled to unset the plugin directory on processor pods through top-level extraEnv.

charts/influxdb3-enterprise/README.md
alespour: addressed in https://github.com/influxdata/helm-charts/pull/812/commits/dabb74fa9fdecaf14d96e6441c711bcb327f4ad6

bednar (approved):
Looks good. The follow-up docs now clarify the top-level extraEnv behavior when processor pods are enabled, and the rendered scenarios/checks I verified pass.

Issue #814, opened 2026-07-15
influxdb3-enterprise: license.type: home is incompatible with the chart's multi-node architecture


matthiasnickac:
## Summary

`license.type: home` is offered as a documented value in `values.yaml`, but the chart's architecture makes it impossible to actually run under a Home license: every component (`ingester`, `querier`, `compactor`, `processor`) registers as a separate node, and the Home license's license server enforces a hard single-node limit at runtime.

## Environment

- OS / Kubernetes distribution: Talos Linux
- Kubernetes version: server `v1.32.2` (client `v1.34.1`)
- Helm chart: `influxdb3-enterprise` version `0.7.3`, appVersion `3.9.3`
- Deployment: standard `helm install`, `ingester.replicas: 1`, `querier.replicas: 1`, default compactor
- License: `license.type: "home"`, verified via email

## Steps to reproduce

1. Install the chart with `license.type: "home"` and a verified license email
2. Leave `ingester`, `querier`, and `compactor` at their default enabled state (1 replica each)
3. Observe pod logs

## Expected behavior

Either:
- The deployment starts successfully under a Home license, or
- The chart documents/validates up front that `license.type: home` is incompatible with its default topology, rather than letting it fail at container startup

## Actual behavior

Ingester and compactor pods crash on startup:

```
INFO influxdb3_enterprise_licensing::trial: Home Enterprise license verified
INFO influxdb3_lib::commands::serve: valid license found, happy data crunching
Serve command failed: only one node is allowed when using Home License
```

The license itself validates correctly — the failure is purely about node count.

## Root cause

Each component's `--mode` is a hardcoded literal in its own template, not exposed via `values.yaml`:

```
templates/ingester-statefulset.yaml:            - --mode=ingest
templates/querier-statefulset.yaml:             - --mode=query
templates/compactor-statefulset.yaml:            - --mode=compact
templates/processor-statefulset.yaml:            - --mode=process
```

There's no template or values field that produces a combined `--mode=all` node, which is what a single-node Home license deployment would require. So regardless of how `replicas` or `enabled` are set per component, a Home-license-compatible deployment isn't reachable through this chart's current templates — disabling `querier`/`compactor` doesn't help either, since `ingester` alone can't serve queries.

## Related issues

[...]

PR #815, opened 2026-07-23
fix(influxdb3-ent): reject unsupported configurations

### LINKED ISSUE #809: [influxdb3-enteprise] memory and memory-throttled storage type must not be supported
The chart always splits into separate per-component `StatefulSets` (ingester/querier/compactor/processor), each its own pod, each its own process. In-process memory can't be shared across pods, so there's no shared catalog (because it is object-store backed) → no cluster.

* document it
* adjust storage validation helper

alespour:
Rejects configurations that cannot work in this distributed (InfluxDB multi-node) chart:
- object store type `memory` and `memory-throttled` #809
- `home` license #814

Closes #809 #814

### DIFF
--- a/charts/influxdb3-enterprise/Chart.yaml
+++ b/charts/influxdb3-enterprise/Chart.yaml
@@ -3,7 +3,7 @@ name: influxdb3-enterprise
 description: A Helm chart for deploying InfluxDB 3 Enterprise on Kubernetes
 type: application
 version: 0.8.1
-appVersion: "3.9.9"
+appVersion: "3.9.11"
 keywords:
   - influxdb
   - time-series
--- a/charts/influxdb3-enterprise/Chart.yaml
+++ b/charts/influxdb3-enterprise/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: influxdb3-enterprise
 description: A Helm chart for deploying InfluxDB 3 Enterprise on Kubernetes
 type: application
-version: 0.8.0
-appVersion: "3.9.3"
+version: 0.8.1
+appVersion: "3.9.9"
 keywords:
   - influxdb
   - time-series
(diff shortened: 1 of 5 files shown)


charts/influxdb3-enterprise/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: influxdb3-enterprise
 description: A Helm chart for deploying InfluxDB 3 Enterprise on Kubernetes
 type: application
-version: 0.8.0
-appVersion: "3.9.3"
+version: 0.8.1
+appVersion: "3.9.11"
```
bednar: The changes look good otherwise, but CI should be green before this is merged. The previous commit using 3.9.9 passed both Kubernetes jobs; after this bump to 3.9.11, the Enterprise pods fail to start with `failed to initialize trial license: UnexpectedLicenseResponse(402)`. This may be an external licensing-service issue rather than an image regression, so a successful rerun is sufficient if that is the cause.

bednar (dismissed):
The chart changes for rejecting unsupported object storage and license configurations look good overall. Approving the implementation; please make sure the failing CI install check is resolved or successfully rerun before merging.

alespour:
Postponed after upgrade to 3.10.

charts/influxdb3-enterprise/Chart.yaml
alespour: CI test credentials are good now and CI is 💚 now.

alespour:
on hold - #822 has priority

alespour:
Priority PR #822 was merged. CI is green for this PR.

bednar (approved):
Reviewed the current revision. The validation and documentation updates consistently reject the unsupported multi-node configurations, and the CI checks are green. Approved.

PR #819, opened 2026-08-07
feat(influxdb3-ent): upgrade InfluxDB to 3.10

alespour:
## Summary

Updates the InfluxDB 3 Enterprise from `3.9` to `3.10.5`. InfluxDB 3.10 automatically performs a one-way catalog v2->v3 migration, so this PR adds an upgrade guide, explicit acknowledgement gate.

### Helm users

Back up the catalog by following `UPGRADING-3.9-TO-3.10.md`, then persist `acknowledgeCatalogMigration: true` in the values file. Helm blocks the upgrade until the acknowledgement is set.

### GitOps users

Template-only renderers such as Argo CD do not expose Helm upgrade state, so the chart cannot enforce the gate. Back up the catalog and commit the acknowledgement to the Git-managed values before synchronizing.

## Design notes

- Helm can't detect the previous `appVersion`, so the gate fires on any upgrade
- No rollout ordering; `--atomic` and `helm rollback` unsafe post-migration
- A headless governing Service is the preferred long-term internode design, but adopting it requires recreating the StatefulSet because `spec.serviceName` is immutable. To avoid combining that risk with the one-way catalog migration, this release uses one pod-selecting `ClusterIP` Service per ingester replica. We will address headless discovery in a future chart release; until then, ingesters should be scaled through Helm.

### DIFF
--- a/charts/influxdb3-enterprise/Chart.yaml
+++ b/charts/influxdb3-enterprise/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: influxdb3-enterprise
 description: A Helm chart for deploying InfluxDB 3 Enterprise on Kubernetes
 type: application
-version: 0.8.0
-appVersion: "3.9.3"
+version: 0.9.0
+appVersion: "3.10.5"
 keywords:
   - influxdb
   - time-series
--- a/charts/influxdb3-enterprise/UPGRADING-3.9-TO-3.10.md
+++ b/charts/influxdb3-enterprise/UPGRADING-3.9-TO-3.10.md
@@ -1,0 +1,75 @@
+# Upgrade from InfluxDB 3 Enterprise 3.9 to 3.10
+
+InfluxDB 3.10 automatically, idempotently, and crash-safely migrates catalog
+v2 to v3 on first startup. The migration is one-way: InfluxDB 3.9 cannot read
+the migrated catalog. Returning to 3.9 requires restoring a pre-upgrade catalog
+backup.
+
+See the official [upgrade](https://docs.influxdata.com/influxdb3/enterprise/admin/upgrade/),
+[backup](https://docs.influxdata.com/influxdb3/enterprise/admin/backup-restore/),
+and [release](https://docs.influxdata.com/influxdb3/enterprise/release-notes/)
+documentation.
+
+## Before upgrading
+
+```bash
+export RELEASE=influxdb3-enterprise
+export NAMESPACE=influxdb3
+export VALUES_FILE=./my-values.yaml
+export TARGET_CHART_VERSION=0.9.0
+
+helm get values "$RELEASE" -n "$NAMESPACE" > pre-3.10-values.yaml
+kubectl get pods -n "$NAMESPACE" \
+  -l "app.kubernetes.io/instance=$RELEASE" \
+  -o custom-columns=NAME:.metadata.name,IMAGE:.spec.containers[0].image,READY:.status.containerStatuses[0].ready
+```
+
+Quiesce writes. Back up `{prefix}/catalog/v2/logs/` and
+`{prefix}/catalog/v2/snapshot` in object storage, and verify that the backup can
+be restored.
+
+If `VALUES_FILE` sets `image.tag`, remove the override or change it to
+`3.10.5-enterprise`. Otherwise, the chart continues deploying the overridden
+image.
+
+## Upgrade
+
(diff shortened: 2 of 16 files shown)


charts/influxdb3-enterprise/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: influxdb3-enterprise
 description: A Helm chart for deploying InfluxDB 3 Enterprise on Kubernetes
 type: application
-version: 0.8.0
-appVersion: "3.9.3"
+version: 0.9.0
+appVersion: "3.10.5"
```
bednar, line 6: [P1] Wire the 3.10 internode gRPC path before changing the app version. InfluxDB 3.10 routes Processing Engine writes from non-ingester nodes through internode gRPC and requires `--conn-info` together with `--internode-bind-addr`. The processor examples in this chart call `influxdb3_local.write`, but the rendered ingester and processor StatefulSets and Services expose only HTTP/8181 and set neither option, so enabling `processingEngine` with this image breaks those writes. Please add the internode listener, Service and NetworkPolicy wiring, point processor nodes at the ingester gRPC endpoint, and cover the rendered configuration with a test.

charts/influxdb3-enterprise/UPGRADING-3.9-TO-3.10.md
```diff
@@ -0,0 +1,79 @@
...
+  -n "$NAMESPACE" \
+  --version "$TARGET_CHART_VERSION" \
+  -f "$VALUES_FILE" \
+  --wait --timeout 30m
+```
+
+Do not use `--atomic`, and do not run `helm rollback` to a 3.9 revision after
+the catalog migration. Both can restore a 3.9 image while leaving the catalog
+at v3. To return to 3.9, first restore the pre-upgrade catalog backup. Do not
+enable `INFLUXDB3_UPGRADE_PACHA_TREE`; migrate the storage engine separately.
```
bednar: [P2] Name the actual 3.10 storage-engine switch here. `INFLUXDB3_UPGRADE_PACHA_TREE` is not a documented 3.10 setting; the supported switch is `--use-pacha-tree` or `INFLUXDB3_ENTERPRISE_USE_PACHA_TREE`. As written, an operator who already sets the real option may not recognize that this warning applies and may combine the catalog and storage-engine migrations. Please replace this name with the supported flag and environment variable.

bednar (changes requested):
Two issues need to be addressed before this version bump is safe.

charts/influxdb3-enterprise/UPGRADING-3.9-TO-3.10.md
alespour: addressed in https://github.com/influxdata/helm-charts/pull/819/commits/6f3278c52c2e81874a1246ddc82abb7439922977

charts/influxdb3-enterprise/Chart.yaml
alespour: Missing gRPC wiring added.

charts/influxdb3-enterprise/templates/ingester-statefulset.yaml
```diff
@@ -59,6 +59,10 @@ spec:
             - serve
             - --mode=ingest
             - --node-id=$(POD_NAME)
+            {{- if .Values.processingEngine.enabled }}
+            - --internode-bind-addr=0.0.0.0:{{ include "influxdb3-enterprise.ingesterInternodePort" . }}
+            - --conn-info=$(POD_NAME).{{ include "influxdb3-enterprise.fullname" . }}-ingester:{{ include "influxdb3-enterprise.ingesterInternodePort" . }}
```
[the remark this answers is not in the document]
alespour: @copilot The portability concern is valid: the Kubernetes DNS specification only guarantees `<hostname>.<service>` endpoint records for headless Services. However, changing this `StatefulSet` to a new headless governing Service is not upgrade-safe for this chart release because `spec.serviceName` is immutable on existing `StatefulSet`s. The current CoreDNS implementation does resolve the advertised name (and the Helm hook passes), so the statement that the hook will fail in current CI is not accurate, but that test does not prove portability to other DNS implementations. We are intentionally retaining the existing governing `Service` for this upgrade and deferring that service-topology migration to a future breaking chart release.

bednar (approved):
The requested changes have been addressed. The updated internode gRPC wiring, migration documentation, and verification coverage look good.

PR #822, opened 2026-08-17
fix(influxdb3-ent): make catalog migration acknowledgement one-time

alespour:
## Summary

Fixes the InfluxDB 3.10 catalog-migration acknowledgement introduced in chart 0.9.0, which otherwise remained required for every future Helm upgrade.

  - Records `influxdata.com/catalog-format: v3` on the release `ConfigMap` for recognized InfluxDB 3.10+ images and preserves an existing marker during subsequent live upgrades.
  - Uses Helm lookup during live upgrades to skip the acknowledgement once the marker exists.

## User impact

- 0.8.x users must back up the catalog and acknowledge the one-way migration once.
- 0.9.0 users receive the missing marker during their 0.9.1 upgrade.
- Later live upgrades no longer require the acknowledgement.
- Client-side renders still require the flag because they cannot query the live marker.
- The supported migration path uses the chart’s default image. Users overriding `image.tag` are responsible for ensuring the selected image is compatible with the catalog format.

## Verification

- helm lint passed.
- Default and acknowledgement render paths passed.
- Missing-marker upgrades were blocked as expected.
- Marker emission was verified for supported image tags.
- Existing-marker preservation was verified when the target image would not independently emit the marker.
- Live server-side lookup was verified with the marker present and absent.

### DIFF
--- a/charts/influxdb3-enterprise/templates/configmap.yaml
+++ b/charts/influxdb3-enterprise/templates/configmap.yaml
@@ -6,6 +6,8 @@
 {{- include "influxdb3-enterprise.validateObjectStoreTlsCa" . }}
 {{- include "influxdb3-enterprise.validateAdminTokenConfig" . }}
 {{- include "influxdb3-enterprise.validatePermissionTokensConfig" . }}
+{{- $existing := lookup "v1" "ConfigMap" .Release.Namespace (include "influxdb3-enterprise.configMapName" .) | default dict }}
+{{- $recordedFormat := dig "metadata" "annotations" "influxdata.com/catalog-format" "" $existing }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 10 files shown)


charts/influxdb3-enterprise/templates/configmap.yaml
```diff
@@ -9,10 +9,14 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ include "influxdb3-enterprise.fullname" . }}-config
+  name: {{ include "influxdb3-enterprise.configMapName" . }}
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "influxdb3-enterprise.labels" . | nindent 4 }}
+  {{- if eq (include "influxdb3-enterprise.targetUsesCatalogV3" .) "true" }}
```
bednar: Once this chart has written the catalog marker, a later upgrade with a pre-3.10 or otherwise unrecognized image tag renders this ConfigMap without the annotation. Helm will then remove the existing `influxdata.com/catalog-format: v3` annotation, so the next upgrade will fail the lookup and require the acknowledgement again. That breaks the promised one-time behavior. Please preserve an existing v3 marker instead of deleting it when the current tag is not recognized (and add a regression test for this transition).

bednar (changes requested):
Please preserve the catalog marker after it has been recorded; otherwise later upgrades can require the acknowledgement again.

charts/influxdb3-enterprise/templates/configmap.yaml
alespour: fixed in https://github.com/influxdata/helm-charts/pull/822/commits/7d10206f11f0f8d780f80a4332f49236df896fec

The fix verified locally. Current CI is not prepared for testing upgrades, we would have to add specific step for such test.

bednar (approved):
Reviewed the current patch; no further changes requested.

PR #824, opened 2026-08-25
feat(telegraf-ds|telegraf): add annotations for the workload object

ranyhb:
Why: both charts let you annotate the pods (podAnnotations) but not the DaemonSet/Deployment object itself. We need pulumi.com/skipAwait on the object and currently post-process the rendered manifest to get it.

How: adds a top-level annotations value on the workload object's metadata, guarded with {{- with }}, defaults to {}. Supports tpl like podAnnotations does since #777. Both charts changed together (same as #777); versions bumped 1.1.53→1.1.54 and 1.8.74→1.8.75.

Testing: with the version held equal, default render is byte-identical on both charts (211 / 186 lines). With the value set, only the object's metadata.annotations appears — verified via the new ci/workload-annotations-values.yaml, including tpl expansion. helm lint passes on both.

- [ ] CHANGELOG.md updated ( nothing to udpate - no changelog.md) 
- [X] Rebased/mergable
- [X] Tests pass
- [X] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

### DIFF
--- a/charts/telegraf-ds/Chart.yaml
+++ b/charts/telegraf-ds/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: telegraf-ds
-version: 1.1.53
+version: 1.1.54
 appVersion: 1.39.3
 deprecated: false
 description: Telegraf is an agent written in Go for collecting, processing, aggregating, and writing metrics.
--- a/charts/telegraf-ds/ci/workload-annotations-values.yaml
+++ b/charts/telegraf-ds/ci/workload-annotations-values.yaml
@@ -1,0 +1,3 @@
+annotations:
+  pulumi.com/skipAwait: "true"
+  example.com/release: "{{ .Release.Name }}"
--- a/charts/telegraf-ds/ci/workload-annotations-values.yaml
+++ b/charts/telegraf-ds/ci/workload-annotations-values.yaml
@@ -1,3 +1,3 @@
-annotations:
+daemonSetAnnotations:
   pulumi.com/skipAwait: "true"
   example.com/release: "{{ .Release.Name }}"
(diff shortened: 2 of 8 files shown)


ranyhb:
@skartikey @alespour @srebhan — could one of you take a look when you have a moment? No CODEOWNERS on the repo, and @skartikey merged #777 which this follows.

**TL;DR:** both charts expose `podAnnotations` (annotates the pods) but there's no way to annotate the DaemonSet/Deployment object itself. Adds a top-level `annotations` value, `{}` by default, guarded with `with` — default render is byte-identical. Supports `tpl` like `podAnnotations` does since #777, and both charts are changed together the same way #777 did.

ranyhb:
@alespour @skartikey @srebhan — mind taking a look? Small PR (+23/-2).

Adds a top-level `annotations` value to `telegraf-ds` and `telegraf` so the DaemonSet/Deployment object can be annotated (`podAnnotations` already covers the pods). Defaults to `{}`, byte-identical render when unset, follows #777 including `tpl` support.

CLA is signed. One thing that needs a maintainer: CI hasn't run — the workflow needs an "Approve and run workflows" click since this is my first PR here.

skartikey:
@ranyhb Thanks, this looks good and renders clean. One ask before merge: could you rename the value to `deploymentAnnotations` (telegraf) and `daemonSetAnnotations` (telegraf-ds)? Every other annotations value in this repo carries its target object in the name, either nested like `service.annotations` or prefixed like `podAnnotations`, and a bare top-level `annotations` is easy to misread as "stamp every object" the way `commonAnnotations` works in other chart ecosystems. It's a cheap rename now but a breaking values change once released.

ranyhb:
@skartikey Hi. thanks for reviewing. on it.

ranyhb:
@skartikey renamed to `daemonSetAnnotations`/`deploymentAnnotations` as asked — what do you think?

skartikey:
@ranyhb Rename looks good, thanks. Rendered both charts against master with the version held equal and the default output is byte-identical, and the new ci values expand tpl as expected. Approving; I have kicked off the workflows and will merge once CI is green.

skartikey (approved):

PR #825, opened 2026-08-25
feat(influxdb3-ent): upgrade InfluxDB to 3.11

alespour:
This PR updates the InfluxDB 3 Enterprise to 3.11.2.

It takes the upgrade and compatibility subset of #818, leaving larger additive features for separate work. Existing Helm values remain compatible, with preferred 3.11 names added where needed.

`PachaTree` migration is intentionally a separate operator step. The chart should first be upgraded to 3.11.2 and verified,  then migration started with:

```bash
  helm upgrade <release> influxdata/influxdb3-enterprise \
    --reuse-values \
    --set acknowledgePachaTreeMigration=true
```

The acknowledgement maps to the officially documented [`INFLUXDB3_UPGRADE_PACHA_TREE=true`](https://docs.influxdata.com/influxdb3/enterprise/reference/internals/storage-engine/#run-the-storage-engine-upgrade) trigger; `helm upgrade --set` is the chart-level mechanism for restarting the cluster with it enabled.

Enabling migration during the initial 3.11 rollout is rejected (by InfluxDB binary) because the Parquet compactor must run first. The upgrade guide documents the required sequence, verification, and downgrade limitations.


charts/influxdb3-enterprise/tests/manual/scenarios/upgrade-3.10-to-3.11.sh
```diff
@@ -0,0 +1,311 @@
...
+
+  while [ "$SECONDS" -lt "$deadline" ]; do
+    if status_output=$("${kubectl_command[@]}" exec \
+      --namespace "$namespace" "$ingester_pod" -- \
+      influxdb3 query \
+      --host "$querier_host" \
+      --database _internal \
+      --token "$auth_token" \
+      --format csv \
+      'SELECT status FROM system.upgrade_parquet_node'); then
```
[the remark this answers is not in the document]
alespour: `system.upgrade_parquet_node` is the documented completion check: the official docs say to wait until every node reports `completed`. `system.upgrade_parquet` provides per-file progress and may contain valid terminal statuses  such as `skipped_source_missing` or `skipped_table_dropped`, so requiring every file row to equal `completed` would incorrectly fail a completed migration.

https://docs.influxdata.com/influxdb3/enterprise/admin/query-system-data/

Rejected.

charts/influxdb3-enterprise/tests/manual/scenarios/upgrade-3.10-to-3.11.sh
```diff
@@ -0,0 +1,311 @@
...
+  $1 == "secretAccessKey:" {
+    gsub(/"/, "", $2)
+    print $2
+    exit
+  }
+' "$values_file")
+
+export AWS_ACCESS_KEY_ID=$aws_access_key
+export AWS_SECRET_ACCESS_KEY=$aws_secret_key
+
```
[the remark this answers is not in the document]
actuallysourab: You're right, and Copilot flagged the same thing — you're both reviewing `b80881a`, before the fix landed in `d4bb148`. The wording no longer claims "earliest".

Your diagnosis is the more precise one, so for the record: `order_by_soft_preferences` reranks, `ordering._interleave_by_day` emits the earliest slot of each **DAY** in turn, and `proposal.candidates[:OPTIONS_CAP]` truncates — so a widened week offers Mon/Tue/Wed 09:00 while the three genuinely earliest all sit on Monday. Confirmed against `_interleave_by_day`'s own docstring.

I took your **first** option — describe them as alternatives, no adjective about when. It now reads *"but here are some other times that could work instead:"*

I did **not** take the second (sort chronologically before the cap), and want to be explicit about why: `spread_across_days` exists precisely to stop the cap spending itself on the earliest day. TMASA-138 fixed exactly that — five Monday-morning times in answer to "what times are you available this week?" Sorting chronologically before the cap would reintroduce that bug to make a sentence true, when changing the sentence costs nothing. The day-spread is also what makes the offer useful on a booked-solid calendar, which is the whole point of TMASA-193.

Also fixed the thing your comment pointed at that I'd missed: **`shared-contracts.md:629` quoted the old wording verbatim** — drift I introduced by changing the constant without the doc. Corrected in `297a3ab`, and the registry now records *why* the sentence carries no adjective, plus that its length is load-bearing, so it doesn't get "improved" back.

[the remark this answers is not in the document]
alespour: fixed in https://github.com/influxdata/helm-charts/pull/825/commits/bffd960f08332752bdf0a607f69d6c07af3cf4e1

charts/influxdb3-enterprise/tests/manual/scenarios/upgrade-3.10-to-3.11.sh
```diff
@@ -0,0 +1,353 @@
...
+  exit "$status"
+}
+
+trap cleanup EXIT
+trap 'exit 130' INT TERM
+
+log "Creating temporary S3 bucket: $bucket"
+aws --endpoint-url "$s3_endpoint" \
+  s3api create-bucket --bucket "$bucket" --no-cli-pager
```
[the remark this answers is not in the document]
alespour: addressed in https://github.com/influxdata/helm-charts/pull/825/commits/fa4416392f0327a844bfd4deee51170cadb3b242

charts/influxdb3-enterprise/UPGRADING-3.10-TO-3.11.md
```diff
@@ -0,0 +1,144 @@
...
+`--set acknowledgeCatalogMigration=true` to the version upgrade. Keep
+`acknowledgePachaTreeMigration=false` until the catalog migration and the
+3.11.2 deployment have both been verified.
+
+Catalog migration and PachaTree migration are separate operator decisions and
+use separate acknowledgements.
+
+## Migrate an existing cluster to PachaTree
+
+New clusters created by InfluxDB 3.11 use PachaTree automatically. Clusters
```
bednar: New 3.11 clusters already default to PachaTree, while `acknowledgePachaTreeMigration` only triggers the upgrade for clusters that started on 3.10 or earlier. Please make that scope explicit here, so readers do not infer that this value affects fresh installations.

bednar (changes requested):
Please address the inline documentation clarification before merging.

charts/influxdb3-enterprise/templates/configmap.yaml
```diff
@@ -22,15 +23,24 @@ metadata:
   {{- end }}
 data:
   # Cluster configuration (env-style)
-  INFLUXDB3_ENTERPRISE_CLUSTER_ID: {{ .Values.cluster.id | quote }}
+  INFLUXDB3_CLUSTER_ID: {{ .Values.cluster.id | quote }}
```
bednar, line 26: **[P1] Keep the 3.10 aliases available during the staged rollout**

This ConfigMap is shared by every StatefulSet revision and is updated before a partitioned rollout finishes. If a pod that is still on 3.10.5 is recreated, it reads this updated ConfigMap, but the pre-3.11 binary does not recognize `INFLUXDB3_CLUSTER_ID` or the other new names, so it loses required settings such as the cluster ID and can fail to recover correctly. InfluxDB 3.11 gives the new variable precedence when both are set, so please emit both old and new names for this transition, or use revision-specific ConfigMaps, to keep old pods recoverable during the documented staged upgrade. See the [3.11 compatibility notes](https://docs.influxdata.com/influxdb3/enterprise/reference/config-options/#configuration-name-changes-in-311).

charts/influxdb3-enterprise/tests/manual/scenarios/upgrade-3.10-to-3.11.sh
```diff
@@ -0,0 +1,360 @@
...
+  --namespace "$namespace" \
+  --from-file=admin-token.json="$admin_token_file"
+
+log "Installing chart 0.9.2 / InfluxDB 3.10.5"
+helm install "$release" influxdata/influxdb3-enterprise \
+  "${helm_context[@]}" \
+  --namespace "$namespace" \
+  --version 0.9.2 \
+  --values "$values_file" \
+  --set-string objectStorage.s3.endpoint="$s3_endpoint" \
```
bednar, line 305: **[P2] Use separate host-facing and in-cluster S3 endpoints**

`$s3_endpoint` is used by the host-side AWS CLI above and then injected into the chart here. The bundled `*.svc.cluster.local` value cannot be resolved by the host; overriding it with a localhost port-forward makes it unreachable from the InfluxDB pods. The new flag therefore does not resolve the original reachability problem. Please keep separate endpoints for the host and pods, or perform bucket setup and cleanup from an in-cluster client, and document the invocation.

bednar (changes requested):
Please address the two inline upgrade-safety issues before merging.

charts/influxdb3-enterprise/templates/configmap.yaml
alespour: fixed in https://github.com/influxdata/helm-charts/pull/825/commits/c2b04bb7d5e79262d333223ebf146eb2eda9c107

charts/influxdb3-enterprise/tests/manual/scenarios/upgrade-3.10-to-3.11.sh
alespour: no more host-side AWS CLI required with https://github.com/influxdata/helm-charts/pull/825/commits/5de4f6f215d17143e0988646ad8a5e77a97a4851

charts/influxdb3-enterprise/UPGRADING-3.10-TO-3.11.md
alespour: addressed in https://github.com/influxdata/helm-charts/pull/825/commits/ed4b71e78ccd600d78d8f33d214d55bc1b420970

bednar (dismissed):
The requested upgrade-safety changes have been addressed. The updated chart renders cleanly and CI passes.

alespour:
@bednar The issues found by @dburton-influxdata (thank you!) were pre-existing and have been addressed using suggested fixes:
- misnamed admin token recovery env var - https://github.com/influxdata/helm-charts/pull/825/commits/d14646b3c9b0428247aa6d4adeaa990b7cc1e8bb
- `http.maxRequestSize` corrupted by `int` filter - https://github.com/influxdata/helm-charts/pull/825/commits/7971d0c87be9d654e496562c488f6ac8825c6f37

alespour:
@bednar `pachaTree` and new processing engine options turned out to be smaller change then I assumed before, and is more logical here. Hence added via https://github.com/influxdata/helm-charts/pull/825/commits/9a65a7b5efe5b734546d8095431b05e62d3f428c.

Now upgrade to 3.11 is complete with this single PR. UI Explorer is a optional feature better to be added in separate PR, because it introduces mode/topology decisions, secret management, and ingress exposure rather than simple configuration variables.

---

The public storage-engine documentation still lists the CLI alias `--replica-gen0-load-concurrency`, which implies the deprecated environment variable `INFLUXDB3_REPLICA_GEN0_LOAD_CONCURRENCY`.

Verified the canonical name directly against the `influxdb:3.11.2-enterprise` binary using `influxdb3 serve --help-all`:

```text
--replica-snapshot-manifest-load-concurrency <N>
  [aliases: replica-gen0-load-concurrency]
  [env: INFLUXDB3_REPLICA_SNAPSHOT_MANIFEST_LOAD_CONCURRENCY=]
```

dburton-influxdata:
will these be published soon for customer testing?

alespour:
> will these be published soon for customer testing?

I think its GTG as soon as @bednar approves it.

bednar (approved):
The follow-up fixes and the additional InfluxDB 3.11 runtime tuning options have been reviewed. The rendered configuration matches the 3.10.5/3.11.2 compatibility requirements, Helm lint passes, and CI is green.

PR #826, opened 2026-08-26
feat(influxdb3-core): upgrade InfluxDB to 3.10.5

alespour:
- upgrades InfluxDB to 3.10.5
- makes SA token automount configurable => closes applicable parity gap with Enterprise chart


bednar (approved):
Reviewed the chart version bump and configurable service account token automounting. Helm rendering was verified for unset, false, and true values, including use of an existing ServiceAccount. CI is green.

PR #801, opened 2026-06-18
Fix: Increase startup probe timeout for influxdb3-enterprise 3.2+

dburton-influxdata:
## Description

This PR fixes CrashLoopBackOff issues across **all components** (ingester, querier, compactor, processor) in InfluxDB 3.2+ deployments by increasing the startup probe timeout to accommodate longer initialization times.

Fixes #800

### Chart Version
- Bump version from \

PR #827, opened 2026-08-31
fix(influxdb3-ent): widen startup probe window

### LINKED ISSUE #800: influxdb3-enterprise: Startup probe timeout too short for v3.2+
## Problem

The default startup probe configuration in the influxdb3-enterprise chart is insufficient for InfluxDB 3.2+ deployments, causing **all components** (ingester, querier, compactor, processor) to crash with **CrashLoopBackOff** and exit code 137 (SIGKILL).

### Current Configuration
\\\yaml
probes:
  startup:
    initialDelaySeconds: 10
    periodSeconds: 5
    timeoutSeconds: 5
    failureThreshold: 12
    # Total window: 70 seconds
\\\

The configuration was designed for ~15 second startups (as noted in the comments), but InfluxDB 3.2+ components take approximately **90-120 seconds** to start due to:

1. **Python Processing Engine initialization** (~6 seconds)
2. **WAL recovery operations** 
3. **V2 compaction consumer preloading**
4. **Compactor-specific**: Primary lease acquisition, snapshot discovery, run set index building
5. **Querier-specific**: Cache initialization, connection pool setup

#### Ingester
From production logs:
\\\
startup time: 96731ms address=0.0.0.0:8181
ERROR influxdb3_server: Error serving connection: hyper::Error(IncompleteMessage)
stream closed EOF for influx-enterprise/influxdb3-enterprise-ingester-0
\\\

#### Compactor
From production logs (5000 lines of normal operation):
\\\
2026-06-23T17:48:00.382842Z DEBUG influxdb3_pacha_tree::file_layout::pt_merger: Merge processed 10636 total rows
stream closed EOF for influx-enterprise/influxdb3-enterprise-compactor-0 (influxdb3)
\\\

Both show normal operation followed by abrupt termination - no errors, panics, or crashes in application logs.

### Impact

- **Ingester**: Pods fail to start on fresh deployments, writes unavailable
- **Compactor**: Background compaction fails, storage efficiency degrades
- **Querier**: Query pods restart, degraded query performance
- **Processor**: Processing Engine plugins fail to initialize
- Upgrades to 3.2+ result in cluster-wide instability
- Exit code 137 indicates SIGKILL termination by Kubernetes

### Environment

- Chart version: 0.1.0
- InfluxDB version: 3.2+ (tested with 3.10.0-enterprise)
- Deployment: AWS EKS, us-gov-west-1
- Object store: S3
- Components affected: **ALL** (ingester, querier, compactor, processor)

## Proposed Solution

Increase startup probe timeout to accommodate 90-120 second initialization:

\\\yaml
probes:
  startup:
    initialDelaySeconds: 30
    periodSeconds: 10
    timeoutSeconds: 10
    failureThreshold: 20
    # Total window: 230 seconds (~2.3x safety buffer)
\\\

This maintains appropriate safety margins while preventing premature termination across all components.

## Related Files

- \charts/influxdb3-enterprise/values.yaml\ (lines 611-638) - Global probe configuration
- \charts/influxdb3-enterprise/Chart.yaml\ (version bump needed)

---

**Note**: The probe configuration in \alues.yaml\ is global and applies to all components via the \influxdb3-enterprise.probes\ helper template in \	emplates/_helpers.tpl\.

I have a PR ready with the fix. Would you like me to submit it?

tomklapka:
Widens the startup probe window and documents the probe settings, which the README did not cover at all.

## Why

Nodes replay from object storage on boot, so startup scales with how much data a node reads back. Support cases report startups past 12 minutes, while the window was 155s in practice. A node killed mid-replay restarts and begins the replay again.

## What changes

`failureThreshold` goes from 30 to 184, leaving `initialDelaySeconds: 10` and `periodSeconds: 5` untouched, as #807 did when it last raised the window. That gives `10 + (184 - 1) × 5 = 925s` before the probe triggers termination; the restart itself then follows `terminationGracePeriodSeconds`.

The arithmetic in the old comments was one period optimistic - the window is `initialDelay + (threshold - 1) × period`, so the documented `10s + (30 × 5s) = 160s` was really 155s. Comments and docs now use the correct form.

## Documentation

A Health Probes section covering three things that were not written down anywhere: `helm --wait` defaults to a five-minute timeout, shorter than the startup it waits for; the liveness probe keeps a narrow `3 × 10s` window once the startup probe succeeds; and the troubleshooting note now uses the `Killing` event as the discriminator rather than exit code 137, which is also the OOMKilled signature.

## Testing

`helm lint`, a render of all four CI values files and stock `values.yaml`, and a kind cluster on the CI node image running the full `ct lint` and `ct install` chain. The probe mechanism itself was verified separately: two otherwise identical pods taking 90s to open `/health`, one with a window shorter than the startup and one longer - the first is killed and restarts, the second reaches `Ready`.

Supersedes #801, which proposed 230s before the longer startups were reported.

Closes #800

### DIFF
--- a/charts/influxdb3-enterprise/README.md
+++ b/charts/influxdb3-enterprise/README.md
@@ -310,10 +310,8 @@ narrower window (`3 × 10s`). A node that answers `/health` and then blocks
 during later startup work can still be restarted; raise
 `probes.liveness.failureThreshold` if that happens.
 
-Two alternatives shorten the startup itself rather than tolerating it:
-[Compacted-Data Startup](#compacted-data-startup) bounds the file-index load,
-and `ingester.persistence` keeps the WAL on a local volume instead of replaying
-it from object storage.
+This widens the window rather than shortening the startup. If nodes routinely
+need most of it, the boot work itself is worth investigating.
 
 #### TLS
 
@@ -640,17 +638,21 @@ kubectl describe pod -n influxdb3 influxdb3-enterprise-ingester-0
 ```
 
 A pod that restarts during startup while its logs show normal activity was
-either killed by the startup probe or ran out of memory. Both exit with code
-137, so check which before changing anything:
+either killed by the startup probe or ran out of memory. Read the termination
+reason and the events rather than the exit code:
 
 ```bash
 kubectl get pod -n influxdb3 influxdb3-enterprise-ingester-0 \
   -o jsonpath='{.status.containerStatuses[0].lastState.terminated.reason}'
+kubectl describe pod -n influxdb3 influxdb3-enterprise-ingester-0 | grep -i probe
 ```
 
 `OOMKilled` means raise the memory limit. `Error` together with a
 `Startup probe failed` event means the probe window is too short; see
-[Health Probes](#health-probes).
+[Health Probes](#health-probes). Exit code 137 is one possible signature of a
+probe kill, not proof of one: kubelet asks the runtime to terminate first and
+honours `terminationGracePeriodSeconds`, so a process that exits during that
+window reports a different code, and 137 is also what an OOM kill produces.
 
 #### License Issues
 
--- a/charts/influxdb3-enterprise/README.md
+++ b/charts/influxdb3-enterprise/README.md
@@ -282,6 +282,39 @@ ingester:
     numThreads: 20
 ```
 
+#### Health Probes
+
+All components share one probe configuration. The startup probe guards the
+initialization window; liveness and readiness start only once it succeeds.
+
+```yaml
+probes:
+  startup:
+    initialDelaySeconds: 10
+    periodSeconds: 5
+    timeoutSeconds: 5
+    failureThreshold: 184   # 10s + (184 - 1) × 5s = 925s
+```
+
+Nodes replay from object storage on boot, so startup scales with how much data a
+node reads back, and clusters with a large history have been reported to need
+more than 12 minutes. Two things follow from a window this wide.
+
+`helm install` and `helm upgrade` with `--wait` default to a five-minute
+timeout, which is shorter than the startup they are waiting for. Pass
+`--timeout 20m` to match, or the release fails while the pods are still booting
+normally.
+
+Once the startup probe succeeds, the liveness probe takes over with a much
+narrower window (`3 × 10s`). A node that answers `/health` and then blocks
+during later startup work can still be restarted; raise
+`probes.liveness.failureThreshold` if that happens.
+
+Two alternatives shorten the startup itself rather than tolerating it:
+[Compacted-Data Startup](#compacted-data-startup) bounds the file-index load,
+and `ingester.persistence` keeps the WAL on a local volume instead of replaying
+it from object storage.
+
 #### TLS
 
 Enable TLS with inline cert/key or an existing secret:
@@ -606,6 +639,19 @@ Check events:
 kubectl describe pod -n influxdb3 influxdb3-enterprise-ingester-0
 ```
 
+A pod that restarts during startup while its logs show normal activity was
+either killed by the startup probe or ran out of memory. Both exit with code
+137, so check which before changing anything:
+
+```bash
+kubectl get pod -n influxdb3 influxdb3-enterprise-ingester-0 \
+  -o jsonpath='{.status.containerStatuses[0].lastState.terminated.reason}'
+```
+
+`OOMKilled` means raise the memory limit. `Error` together with a
+`Startup probe failed` event means the probe window is too short; see
+[Health Probes](#health-probes).
+
 #### License Issues
 
 Verify license configuration:
@@ -680,6 +726,12 @@ logs:
 | `security.auth.adminToken.recovery.httpBind` | Bind address for admin token recovery endpoint (`INFLUXDB3_ADMIN_TOKEN_RECOVERY_HTTP_BIND_ADDR`) | `""` |
 | `serviceAccount.automountServiceAccountToken` | Configure automatic mounting of the Kubernetes service account token in component pods and the created ServiceAccount | `not set` |
 | `extraEnv` | Extra environment variables applied to all components | `[]` |
+| `probes.enabled` | Enable liveness, readiness, and startup probes on all components | `true` |
+| `probes.startup.initialDelaySeconds` | Delay before the first startup check | `10` |
+| `probes.startup.periodSeconds` | Interval between startup checks | `5` |
+| `probes.startup.timeoutSeconds` | Timeout of a single startup check | `5` |
+| `probes.startup.failureThreshold` | Failed startup checks before the pod is killed | `184` |
+| `probes.liveness.*` / `probes.readiness.*` | Liveness and readiness timings; see `values.yaml` | see `values.yaml` |
 
 ### Object Storage Parameters
 
--- a/charts/influxdb3-enterprise/README.md
+++ b/charts/influxdb3-enterprise/README.md
@@ -293,7 +293,7 @@ probes:
     initialDelaySeconds: 10
     periodSeconds: 5
     timeoutSeconds: 5
-    failureThreshold: 184   # 10s + (184 - 1) × 5s = 925s
+    failureThreshold: 184   # 10s + (184 - 1) × 5s = 925s to termination
 ```
 
 Nodes replay from object storage on boot, so startup scales with how much data a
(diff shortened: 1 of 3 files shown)


tomklapka:
Verified locally on a kind cluster using the same node image as CI (v1.31.14), running `ct lint`, `ct install` and the workflow's log pattern check. All four CI values files install, the chart's helm test passes, and the pattern check finds nothing.

On the point raised in #800, that probe config cannot affect this much: both readings hold, they just answer different questions. The window does not shorten the replay, and the pod is not `Ready` until the replay finishes. What the window decides is whether the pod is allowed to get that far.

Two otherwise identical pods, each taking 90s before opening `/health`:

| startup window | outcome | restarts |
|---|---|---|
| 40s, shorter than the start | killed, `exitCode: 137` | 1, then repeating |
| 310s, longer than the start | `Ready` | 0 |

For the first one the kubelet logs `Container app failed startup probe, will be restarted`, while the application log shows only its normal startup work and no error. That is the signature #800 describes: exit 137 with clean application logs. Each restart then begins the replay from the start, which is the loop being reported.

So a wider window does not make a slow start faster, and this PR does not claim that it does. It stops a slow start from turning into a restart loop.

One tradeoff worth stating: a wider window also delays the point at which a pod that will never start gets killed. In CI that is hidden behind the `helm install --wait` timeout, which fires first.

tomklapka:
Updated after a review pass. Four things changed.

The window now moves through `failureThreshold` alone (12 -> 184), leaving `initialDelaySeconds` and `periodSeconds` as they were. Raising those delayed every normal startup by 20-30s without widening anything, and #807 set the precedent of touching only the threshold.

The arithmetic was one period optimistic. The window is `initialDelay + (threshold - 1) x period`, so the old `10s + (30 x 5s) = 160s` comment described 155s. Comments and docs now use the correct form.

Exit code 137 is `128 + SIGKILL`, which is also the OOMKilled signature, so the troubleshooting note no longer treats it as proof of a probe kill. It now shows how to read `lastState.terminated.reason` and distinguish the two before touching any threshold.

Three gaps are documented rather than silently left open. `helm --wait` defaults to a five-minute timeout, shorter than the startup it is waiting for, so the section says to pass `--timeout 20m`. The liveness probe keeps its narrow `3 x 10s` window once the startup probe succeeds, which can still restart a node that answers `/health` and then blocks. And both `compactedData` (#828) and `ingester.persistence` shorten the startup itself instead of tolerating it, which is the better fix where it applies.

Not addressed here, worth separate issues if wanted: the probe block is shared by all four components, so a misconfigured querier now takes 15 minutes to surface instead of 160s, and `influxdb3-core` still carries the original 70s window with a comment describing a 15-second startup.

charts/influxdb3-enterprise/README.md
```diff
@@ -282,6 +282,39 @@ ingester:
...
+normally.
+
+Once the startup probe succeeds, the liveness probe takes over with a much
+narrower window (`3 × 10s`). A node that answers `/health` and then blocks
+during later startup work can still be restarted; raise
+`probes.liveness.failureThreshold` if that happens.
+
+Two alternatives shorten the startup itself rather than tolerating it:
+[Compacted-Data Startup](#compacted-data-startup) bounds the file-index load,
+and `ingester.persistence` keeps the WAL on a local volume instead of replaying
```
bednar: This recommendation is not actionable in this chart version. The `Compacted-Data Startup` anchor and its values do not exist here yet (they are being added separately in #828), while `ingester.persistence` is a deprecated compatibility value that no template consumes; this README also states below that the old WAL PVCs were unused and the WAL is persisted in object storage. Please remove these alternatives from this PR, or document only mechanisms that are actually available in the released chart.

charts/influxdb3-enterprise/README.md
```diff
@@ -606,6 +639,19 @@ Check events:
 kubectl describe pod -n influxdb3 influxdb3-enterprise-ingester-0
 ```
 
+A pod that restarts during startup while its logs show normal activity was
+either killed by the startup probe or ran out of memory. Both exit with code
+137, so check which before changing anything:
```
bednar: Please do not state that a startup-probe restart always exits with 137. Kubelet honors `terminationGracePeriodSeconds` for failed startup/liveness probes and first asks the runtime to terminate the container; 137 occurs only when the process is ultimately killed with SIGKILL. A process that exits during graceful termination can therefore have a different code. Please phrase this as exit 137 being one possible signature and keep the termination reason plus probe event as the actual discriminator.

bednar (changes requested):
Two documentation issues need correction before merge; both comments are inline.

charts/influxdb3-enterprise/README.md
tomklapka: Removed both. Verified your point about `ingester.persistence`: no template references it, and `values.yaml` marks it as retained only so older values files stay recognisable. The compacted-data anchor was a forward reference to #828 and does not resolve here.

The paragraph now just notes that this widens the window rather than shortening the startup, without naming remedies this chart version lacks.

charts/influxdb3-enterprise/README.md
tomklapka: Reworded. The section now leads with the termination reason and the probe event as the discriminator, and adds a `describe | grep -i probe` alongside the jsonpath. Exit 137 is described as one possible signature, with the note that kubelet asks the runtime to terminate first and honours `terminationGracePeriodSeconds`, so a process exiting in that window reports a different code - and that 137 is also the OOM signature.

My local repro did produce 137, but the test container ignored SIGTERM, so that was an artefact of the probe rather than evidence about the general case.

charts/influxdb3-enterprise/README.md
```diff
@@ -606,6 +637,23 @@ Check events:
...
+reason and the events rather than the exit code:
+
+```bash
+kubectl get pod -n influxdb3 influxdb3-enterprise-ingester-0 \
+  -o jsonpath='{.status.containerStatuses[0].lastState.terminated.reason}'
+kubectl describe pod -n influxdb3 influxdb3-enterprise-ingester-0 | grep -i probe
+```
+
+`OOMKilled` means raise the memory limit. `Error` together with a
+`Startup probe failed` event means the probe window is too short; see
```
bednar: A generic `Startup probe failed` event does not establish that the probe caused this restart. Kubelet records that warning for every failed probe attempt, including attempts below `failureThreshold`; an unrelated process exit can therefore leave both `Error` and an earlier warning. Please use the `Killing` event whose message says `failed startup probe, will be restarted` as the discriminator. The preceding `either ... or` should also be softened to describe these as common causes rather than the only possible causes of an externally killed container.

charts/influxdb3-enterprise/values.yaml
```diff
@@ -731,13 +731,13 @@ probes:
 
   # Startup probe settings
   # Protects container during initialization. Once successful, never runs again.
-  # Allows slower startup without delaying successful readiness detection.
-  # 10s initial + (30 × 5s) = 160s total.
+  # 10s initial + (184 - 1) × 5s = 925s before the container is restarted.
```
bednar: 925 seconds is the nominal point at which the failed startup probe triggers termination, not necessarily when the container has restarted. As the new README correctly notes, kubelet then honors `terminationGracePeriodSeconds`, so the actual restart can occur later. Please change this wording to something like `before the startup probe triggers termination` so the documented window and graceful-shutdown behavior agree.

bednar (changes requested):
Please also update the PR description before merge. It still describes the obsolete configuration as a 930-second window made from a 30-second initial delay and 90 checks every 10 seconds, while this revision keeps a 10-second initial delay and 5-second period and changes `failureThreshold` from 30 to 184. The description should match the values and timing documented by the final diff. Two additional documentation issues are called out inline.

charts/influxdb3-enterprise/README.md
tomklapka: Fixed. The section now points at the `Killing` event reading `failed startup probe, will be restarted` and notes why `Startup probe failed` is not sufficient, and the either/or is softened to common causes. The window figure in both README and `values.yaml` now reads as the point termination is triggered, with the restart following `terminationGracePeriodSeconds`. PR description updated to the 10s/5s/184 values.

charts/influxdb3-enterprise/values.yaml
tomklapka: Fixed. The section now points at the `Killing` event reading `failed startup probe, will be restarted` and notes why `Startup probe failed` is not sufficient, and the either/or is softened to common causes. The window figure in both README and `values.yaml` now reads as the point termination is triggered, with the restart following `terminationGracePeriodSeconds`. PR description updated to the 10s/5s/184 values.

charts/influxdb3-enterprise/README.md
```diff
@@ -606,6 +637,26 @@ Check events:
 kubectl describe pod -n influxdb3 influxdb3-enterprise-ingester-0
 ```
 
+A pod that restarts during startup while its logs show normal activity is most
```
bednar: A Pod itself does not restart when a startup probe fails; kubelet restarts the affected container within the existing Pod. Please change this to `A container that restarts during startup...` or `A pod whose application container restarts during startup...` so the troubleshooting guidance describes the lifecycle accurately.

charts/influxdb3-enterprise/README.md
```diff
@@ -680,6 +731,12 @@ logs:
 | `security.auth.adminToken.recovery.httpBind` | Bind address for admin token recovery endpoint (`INFLUXDB3_ADMIN_TOKEN_RECOVERY_HTTP_BIND_ADDR`) | `""` |
 | `serviceAccount.automountServiceAccountToken` | Configure automatic mounting of the Kubernetes service account token in component pods and the created ServiceAccount | `not set` |
 | `extraEnv` | Extra environment variables applied to all components | `[]` |
+| `probes.enabled` | Enable liveness, readiness, and startup probes on all components | `true` |
+| `probes.startup.initialDelaySeconds` | Delay before the first startup check | `10` |
+| `probes.startup.periodSeconds` | Interval between startup checks | `5` |
+| `probes.startup.timeoutSeconds` | Timeout of a single startup check | `5` |
+| `probes.startup.failureThreshold` | Failed startup checks before the pod is killed | `184` |
```
bednar: The startup probe does not kill the Pod. After `failureThreshold` consecutive failures, kubelet triggers termination and restart of the affected container while the Pod remains. Please describe this as, for example, `Failed startup checks before container termination is triggered`; that also remains consistent with the `terminationGracePeriodSeconds` explanation above.

bednar (changes requested):
Two Kubernetes object-lifecycle terminology corrections are needed; both comments are inline.

charts/influxdb3-enterprise/README.md
tomklapka: Changed to "A container that restarts during startup", and "killed by" to "terminated by" in the same sentence. a6d551f

charts/influxdb3-enterprise/README.md
tomklapka: Now reads "Failed startup checks before container termination is triggered". The liveness paragraph in Health Probes said the node "can still be restarted" with the same imprecision, so that is now "can still have its container restarted". a6d551f

bednar (approved):
Reviewed the current head. The requested chart and README corrections are addressed, and the checks pass.

PR #828, opened 2026-08-31
feat(influxdb3-ent): expose compacted-data startup options

tomklapka:
Adds `compactedData.skipFileIndex` and `compactedData.loadConcurrencyLimit`.

## Why

Before a node serves traffic it loads compaction state, including a file index mapping indexed column values to Parquet files. On clusters with a large compaction history that load dominates startup: support cases show nodes taking tens of minutes to answer `/health`, and nodes running out of memory before they finish while the application log stays clean.

Both options bound that work and the chart had no way to set them. `skipFileIndex` trades query-time file pruning for a bounded startup footprint; queries stay correct and time-range pruning still applies. It applies to the whole node, so a compactor started with it also skips index merges for new generations. `loadConcurrencyLimit` bounds the memory buffered by in-flight downloads.

This addresses the cause rather than the symptom. #827 widens the startup probe so a slow start is not turned into a restart loop, but the start itself stays slow; these options shorten it.

## Version compatibility

The options predate 3.11 under `INFLUXDB3_ENTERPRISE_COMPACTED_DATA_*` and 3.11 keeps those names as aliases, so the chart emits both spellings. A release therefore works against a pinned 3.9 or 3.10 image as well as 3.11, which matters during the staged rollout described in UPGRADING-3.10-TO-3.11.md.

## Note on ordering with #827

This conflicts with #827 in `Chart.yaml` and `README.md`, since both bump the version and add a section in the same place. They are independent changes; whichever merges second needs a rebase. Suggest merging #827 first as it is the smaller change, and I will rebase this one on top.

## Testing

`helm lint` and a render of all four CI values files plus stock `values.yaml`. `ci/component-extra-env-values.yaml` now exercises both options, so the rendered variable names are covered by `ct install` without adding an install run. Neither option renders when unset, so existing releases are unchanged. Verified against a kind cluster on the CI node image with the full `ct lint` and `ct install` chain.

### DIFF
--- a/charts/influxdb3-enterprise/README.md
+++ b/charts/influxdb3-enterprise/README.md
@@ -100,6 +100,37 @@ At minimum, you must configure:
        host: ""
    ```
 
+### Commercial License
+
+Store the license in a secret and point the chart at it.
+
+1. Create a Kubernetes secret from the license file. The key must be
+   `license-file`:
+   ```bash
+   kubectl -n influxdb3 create secret generic influxdb3-license \
+     --from-file=license-file=/path/to/license.jwt
+   ```
+
+2. Configure the chart:
+   ```yaml
+   license:
+     type: commercial
+     existingSecret: influxdb3-license
+   ```
+
+Notes:
+- Secret key must be `license-file`. The chart mounts that key at
+  `/etc/influxdb/license`. With any other key the mount produces an empty
+  directory and the server logs `failed to read license file from path: Is a
+  directory (os error 21)`.
+- `--from-file=/path/to/license.jwt` without the `license-file=` prefix names
+  the key after the file and hits exactly that case.
+- `license.type` must be `commercial`. It defaults to `trial`, and a trial
+  release reads only `license-email` from the secret and never sets the license
+  file path.
+- The chart does not read a license from object storage. A message about no
+  license found there follows a failed file read; it is not the cause.
+
 ### Preconfigured Admin Token
 
 Use this when you want the cluster to start with a known offline-generated admin token.
@@ -722,6 +753,22 @@ Verify license configuration:
 ```bash
 kubectl get secret -n influxdb3 influxdb3-enterprise-license -o yaml
 ```
+
+`failed to read license file from path: Is a directory (os error 21)` means the
+secret named in `license.existingSecret` has no `license-file` key, so the mount
+produced an empty directory. Check the key name:
+
+```bash
+kubectl -n influxdb3 get secret SECRET_NAME \
+  -o go-template='{{range $k,$v := .data}}{{$k}}{{"\n"}}{{end}}'
+```
+
+That prints key names only. Avoid `-o yaml` or `-o jsonpath='{.data}'` here: both
+print the base64-encoded licence along with the keys.
+
+The `no commercial license found in object store` line that follows is a
+fallback after the failed read, not a separate problem. See
+[Commercial License](#commercial-license).
 
 #### Object Storage Connection
 
--- a/charts/influxdb3-enterprise/README.md
+++ b/charts/influxdb3-enterprise/README.md
@@ -312,6 +312,62 @@ during later startup work can still have its container restarted; raise
 
 This widens the window rather than shortening the startup. If nodes routinely
 need most of it, the boot work itself is worth investigating.
+#### Compacted-Data Startup
+
+Before a node serves traffic it loads compaction state, including a file index
+that maps indexed column values to Parquet files. On clusters with a large
+compaction history that load can dominate startup: nodes take tens of minutes to
+answer `/health`, or run out of memory before they finish.
+
+```yaml
+compactedData:
+  skipFileIndex: true       # skip the in-memory file index
+  loadConcurrencyLimit: 8   # concurrent downloads while loading (server default 20)
+```
+
+`skipFileIndex` trades query-time file pruning for a bounded memory footprint at
+startup. Queries stay correct and time-range pruning still applies, but
+equality-predicate queries scan more files. The setting applies to the whole
+node, so a compactor started with it also skips index merges for newly compacted
+generations. Exempting a single component means overriding both spellings in its
+`extraEnv`, because the ConfigMap supplies both.
+`loadConcurrencyLimit` bounds the memory buffered by in-flight downloads; lower
+it when memory is tight during startup, raise it for faster startup where there
+is network headroom.
+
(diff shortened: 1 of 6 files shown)


tomklapka:
Installed on a kind cluster on the CI node image, which turned up something a render could not.

A fresh 3.11 release defaults to the PachaTree storage engine, and on such a node the server logs:

> these options applied during the storage engine upgrade to PachaTree; the upgrade is complete and they are now ignored — they can be removed from the configuration
> flags=--compacted-data-load-concurrency-limit, --compacted-data-skip-file-index

Both flags are in `MIGRATION_CONSUMED_PARQUET_OPTIONS`, and `upgrade_complete_inert_options` returns them only when the persisted storage mode is fully-adopted PachaTree. During the migration itself, in `Parquet` or `ParquetAndPachaTree` mode, they are live.

That is the right behaviour rather than a problem: the file index is a Parquet mechanism, so the options matter exactly where the failure occurs and are redundant once a cluster has moved past it. But it needed saying, so the README and `values.yaml` now state that they apply to Parquet clusters and to the migration, and that a fully migrated node ignores them.

Full run on the cluster: `ct lint`, `ct install` across all four CI values files, all green. A manual install of the complete four-component cluster confirms both spellings reach every pod's environment and no component complains about an unrecognised variable.

Worth noting for anyone testing this locally: the shared test licence covers 8 cores and each node claims 4 by default, so only two components come up at once and `helm --wait` times out looking like a chart failure. Setting `numCores: 2` on all four fits the whole cluster inside the licence.

charts/influxdb3-enterprise/templates/configmap.yaml
```diff
@@ -197,6 +197,19 @@ data:
   {{- if .tableIndexCacheConcurrencyLimit }}
   INFLUXDB3_TABLE_INDEX_CACHE_CONCURRENCY_LIMIT: {{ .tableIndexCacheConcurrencyLimit | quote }}
   {{- end }}
+
+  {{- end }}
+
+  # Compacted-data startup
+  {{- with .Values.compactedData }}
+  {{- if kindIs "bool" .skipFileIndex }}
```
bednar: kindIs "bool" silently drops quoted values such as skipFileIndex: "true": Helm rendering succeeds, but neither environment variable is emitted, leaving the server default false active. This is particularly risky because operators use this setting to recover nodes that cannot start due to memory pressure. Please use hasKey with ternary, or explicitly fail for non-boolean values, so invalid input is caught during Helm rendering.

bednar (changes requested):
Please address the inline configuration-validation issue before merge.

charts/influxdb3-enterprise/templates/configmap.yaml
tomklapka: Changed to `hasKey` plus an explicit type check, so a quoted or empty value now fails the render:

```
compactedData.skipFileIndex must be an unquoted boolean (true or false), got string: true
```

`kindIs` came from an earlier review pass that flagged `ternary` aborting on a nil with an error naming an unrelated template. Failing explicitly covers both: valid booleans render, anything else stops with a message that names the key.

charts/influxdb3-enterprise/README.md
```diff
@@ -606,13 +694,33 @@ Check events:
...
+`failed to read license file from path: Is a directory (os error 21)` means the
+secret named in `license.existingSecret` has no `license-file` key, so the mount
+produced an empty directory. Check the key name:
+
+```bash
+kubectl -n influxdb3 get secret SECRET_NAME \
+  -o go-template='{{range $k,$v := .data}}{{$k}}{{"\n"}}{{end}}'
+```
+
+That prints key names only. Avoid `-o yaml` or `-o jsonpath='{.data}'` here: both
```
bednar: The command immediately above still tells users to run kubectl get secret ... -o yaml, which this paragraph correctly warns exposes the base64-encoded license. Please replace that command with the key-only go-template using the configured license.existingSecret name, rather than retaining both instructions.

charts/influxdb3-enterprise/README.md
```diff
@@ -606,13 +694,33 @@ Check events:
...
+[Compacted-Data Startup](#compacted-data-startup).
+
 #### License Issues
 
 Verify license configuration:
 ```bash
 kubectl get secret -n influxdb3 influxdb3-enterprise-license -o yaml
 ```
 
+`failed to read license file from path: Is a directory (os error 21)` means the
```
bednar, line 754: Because the license Secret volume is rendered with optional: true, Kubernetes also ignores a Secret that does not exist. This error therefore does not uniquely mean that license-file is missing. Please tell users to verify that license.existingSecret exists in the release namespace as well as checking the key, and mirror that correction in TROUBLESHOOTING.md.

bednar (changes requested):
Please address the two inline license-documentation issues before merge.

charts/influxdb3-enterprise/README.md
tomklapka: Fixed. Verified the volume is rendered with `optional: true`, so a secret that does not exist produces the same empty directory as one without the key - the section now covers both and says a `NotFound` means the secret is missing or in another namespace. TROUBLESHOOTING.md mirrors it. The `-o yaml` command is gone; the key-only go-template serves both checks.

charts/influxdb3-enterprise/README.md
tomklapka: Fixed. Verified the volume is rendered with `optional: true`, so a secret that does not exist produces the same empty directory as one without the key - the section now covers both and says a `NotFound` means the secret is missing or in another namespace. TROUBLESHOOTING.md mirrors it. The `-o yaml` command is gone; the key-only go-template serves both checks.

bednar (dismissed):
The requested changes have been addressed. The updated chart renders correctly and the documentation now covers both missing Secrets and missing license-file keys without exposing Secret values.

tomklapka:
Rebased onto master now that #827 has landed. The version stays at 0.11.0, which is still a valid increment over the 0.10.1 that #827 published.

Three conflicts, all additive and resolved by keeping both sides: the Chart.yaml version line, the new #### section landing at the same anchor as Health Probes, and the parameter table rows. I also added two blank lines the resolution had swallowed, before the Compacted-Data Startup heading and between the two troubleshooting paragraphs.

No change to the chart logic. Re-checked after the rebase that skipFileIndex: false and loadConcurrencyLimit: 0 still emit both variables and that a quoted value still fails the render.

bednar (approved):

PR #829, opened 2026-09-02
docs(influxdb3-ent): document the license secret key

tomklapka:
A customer set `license.existingSecret` and their nodes kept reporting that no license was found. The secret was fine and the licence was valid; its key was not `license-file`.

The chart mounts that specific key at `/etc/influxdb/license` with `optional: true`, so any other key name leaves an empty directory at that path. The server then logs:

```
WARN failed to read license file from path: Is a directory (os error 21)
INFO no commercial license found in object store
```

The second line reads as though the chart looks for the licence in object storage, which sends people looking in the wrong place. It is a fallback after the failed read.

Reproduced on a kind cluster against chart 0.10.0:

| Secret key | `/etc/influxdb/license` | Result |
|---|---|---|
| `license.jwt` | empty directory | `Is a directory (os error 21)` |
| `license-file` | file, 545 B | `Enterprise license verified` |

The required key was documented only in a `values.yaml` comment and one parameter-table row, while the admin token and permission tokens both have a worked example with the key called out. This adds the same for the licence, plus the `license.type: commercial` requirement, since the default `trial` never sets the license file path at all.

No `Chart.yaml` bump. This is documentation, and a version bump on master triggers a release; it can ship with the next one instead.

### DIFF
--- a/charts/influxdb3-enterprise/Chart.yaml
+++ b/charts/influxdb3-enterprise/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: influxdb3-enterprise
 description: A Helm chart for deploying InfluxDB 3 Enterprise on Kubernetes
 type: application
-version: 0.10.0
+version: 0.10.1
 appVersion: "3.11.2"
 keywords:
   - influxdb
--- a/charts/influxdb3-enterprise/README.md
+++ b/charts/influxdb3-enterprise/README.md
@@ -649,8 +649,12 @@ secret named in `license.existingSecret` has no `license-file` key, so the mount
 produced an empty directory. Check the key name:
 
 ```bash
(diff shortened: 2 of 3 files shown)


tomklapka:
Folded into #828 instead of shipping on its own.

CI enforces a `Chart.yaml` bump on every chart change - `chart-testing-action` overrides `CT_CONFIG_DIR`, so the `check-version-increment: false` in our `ct.yaml` never reaches it, which is why this passed `ct lint` locally and failed in Actions. Since a bump on master is also the release trigger, landing this separately would mean cutting a release for a README section. #828 is bumping anyway, so it goes there.

Both review points are carried over: the secret-key check now uses a go-template over `.data` keys rather than `-o jsonpath='{.data}'`, which prints the base64 licence alongside the key names.

Issue #830, opened 2026-09-04
Integrated Enterprise Explorer does not consume documented preconfigured connection config


dburton-influxdata:
## Summary

InfluxDB 3 Enterprise 3.11.2 can serve the integrated Explorer Web UI when querier pods start with `--mode=query,webui` and `--webui-session-secret`, but local testing showed that the integrated Web UI does not appear to consume the documented Explorer preconfigured connection values.

The same documented `config.json` preconfiguration worked when using the standalone Explorer container as a companion Kubernetes Deployment.

This is related to the Explorer follow-up from #817.

## Environment Tested

- InfluxDB 3 Enterprise: `3.11.2`
- Helm chart baseline: released `charts/influxdb3-enterprise` after #825, `appVersion: "3.11.2"`
- Local Kubernetes: Rancher Desktop / k3s on macOS arm64
- Deployment shape: five-node local Enterprise cluster
  - 2 ingesters
  - 2 queriers
  - 1 compactor
- Integrated Explorer endpoint tested through querier Service on port `8181`
- Standalone Explorer companion tested with image `influxdata/influxdb3-ui:1.9.0` on port `8888`

## What Was Configured for Integrated Explorer

Querier pods were configured with:

```text
--mode=query,webui
--webui-session-secret=$(INFLUXDB3_WEBUI_SESSION_SECRET)
```

The documented Explorer preconfiguration was mounted and exposed using both forms:

```text
/app-root/config/config.json
DEFAULT_INFLUX_SERVER
DEFAULT_INFLUX_DATABASE
DEFAULT_API_TOKEN
DEFAULT_SERVER_NAME
```

The mounted `config.json` followed the documented shape:

```json
{
  "DEFAULT_INFLUX_SERVER": "http://127.0.0.1:8181",
  "DEFAULT_INFLUX_DATABASE": "explorer_demo",
  "DEFAULT_API_TOKEN": "<admin-token>",
  "DEFAULT_SERVER_NAME": "Local InfluxDB 3 Enterprise 3.11.2"
}
```

## Expected Behavior

For a fresh browser/session against the integrated Enterprise Explorer, the UI should load or seed the default server connection from `config.json` or the corresponding `DEFAULT_*` environment values.

After session creation, this endpoint should return the configured active server instead of `null`:

```bash
curl -sS -c cookies.txt -b cookies.txt http://localhost:8181/api/sessions
curl -sS -c cookies.txt -b cookies.txt http://localhost:8181/api/server-config/active
[...]

PR #831, opened 2026-09-04
feat(influxdb3-enterprise): add Explorer UI Helm support

dburton-influxdata:
# feat(influxdb3-enterprise): add Explorer UI Helm support

## Summary

Adds chart-level support for InfluxDB 3 Enterprise Explorer with two paths:

- `webui.enabled` enables the integrated Enterprise Web UI on querier pods by rendering `--mode=query,webui` and `--webui-session-secret`.
- `webui.defaultConnection` mounts the documented Explorer `config.json` shape and optional `DEFAULT_*` environment values into querier pods.
- `explorer.enabled` creates an optional standalone Explorer companion Deployment and Service using `influxdata/influxdb3-ui:1.9.0`.
- `explorer.mode` supports `query` or `admin`.
- `explorer.defaultConnection` preconfigures standalone Explorer and defaults the server URL to the in-cluster querier Service.
- Adds validation for required session secrets, invalid Explorer modes, and missing preconfigured tokens.
- Adds NetworkPolicy allowance from standalone Explorer to querier pods.
- Adds README docs plus `examples/values-explorer.yaml` and `ci/explorer-values.yaml`.

## Why

Enterprise 3.11 introduced the integrated Explorer Web UI, but the release notes call out two important deployment requirements:

- the Web UI is not included in plain `--mode all`
- a Web UI session secret is mandatory when `webui` mode is enabled

The standalone Explorer documentation also defines a useful first-run preconfiguration flow through `/app-root/config/config.json` using:

- `DEFAULT_INFLUX_SERVER`
- `DEFAULT_INFLUX_DATABASE`
- `DEFAULT_API_TOKEN`
- `DEFAULT_SERVER_NAME`

## Local Validation

Validation was run against the released 3.11.2 chart baseline and a local five-node 3.11.2 Enterprise cluster.

Commands:

```bash
helm lint charts/influxdb3-enterprise
helm lint charts/influxdb3-enterprise -f charts/influxdb3-enterprise/ci/explorer-values.yaml
helm lint charts/influxdb3-enterprise -f charts/influxdb3-enterprise/examples/values-explorer.yaml

helm template influxdb3-test charts/influxdb3-enterprise \
  --namespace influxdb3-test \
  -f charts/influxdb3-enterprise/ci/explorer-values.yaml
```

Rendered signals checked:

- querier args include `--mode=query,webui`
- querier args include `--webui-session-secret=$(INFLUXDB3_WEBUI_SESSION_SECRET)`
- Web UI default connection config is mounted at `/app-root/config/config.json`
- standalone Explorer renders `influxdata/influxdb3-ui:1.9.0`
- standalone Explorer renders `--mode=admin`
- standalone Explorer config defaults to the in-cluster querier Service URL
- NetworkPolicy allows standalone Explorer to reach querier pods
- missing session secrets and invalid Explorer mode fail at Helm render time

Runtime validation:

- Integrated Enterprise Web UI served successfully from querier pods.
- Integrated Enterprise Web UI accepted mounted/env preconfiguration values, but InfluxDB 3 Enterprise 3.11.2 did not create an active Explorer connection from them during a fresh browser session.
- Standalone Explorer consumed the same documented `config.json` pattern and opened directly into a preconfigured connection in admin mode.
- The preconfigured standalone Explorer session reported InfluxDB 3 Enterprise `3.11.2`, trial license status, and local cluster metadata through the UI.

## Notes

The integrated Web UI and standalone Explorer are disabled by default. Users must explicitly provide a session secret, and preconfigured connections require an API token or an existing Secret.

For a Helm-managed first-run Explorer connection today, the standalone Explorer companion is the verified path. The integrated Web UI chart wiring is still useful for serving the built-in UI from Enterprise, but 3.11.2 may need an upstream Enterprise UI host/WASM change before it consumes the same default connection config.

## Docs

- Enterprise 3.11 integrated Explorer release notes: https://docs.influxdata.com/influxdb3/enterprise/release-notes/#enterprise-2
- Explorer preconfigured connection docs: https://docs.influxdata.com/influxdb3/explorer/install/#pre-configure-influxdb-connections
- Explorer mode docs: https://docs.influxdata.com/influxdb3/explorer/install/#choose-operational-mode


charts/influxdb3-enterprise/values.yaml
```diff
@@ -609,6 +609,92 @@ processingEngine:
...
+  # Optional OpenAI-compatible endpoint for Explorer AI chat.
+  openaiBaseUrl: ""
+
+  # Optional first-connection defaults. The chart mounts config.json using the
+  # documented Explorer shape and exposes DEFAULT_* environment variables.
+  # Note: local validation with Enterprise 3.11.2 showed the standalone
+  # Explorer image consumes this preconfiguration reliably; the integrated
+  # Enterprise Web UI currently accepts the mounted/env values but may still ask
+  # the user to configure the first server.
+  defaultConnection:
```
tomklapka: This block cannot work on 3.11.x and I would leave it out until the host supports it.

The values are delivered correctly - I verified inside a running querier that /app-root/config/config.json is present and readable with the right content, and that all four `DEFAULT_*` variables plus `CONFIG_PATH` are in the container environment. `GET /api/server-config` still returns `[]` and `/api/server-config/active` returns `null`, and the server log never mentions config.json, /app-root or DEFAULT_ during startup. The reason is in `build_wasi_ctx` (#830).

There is a cost beyond the option doing nothing: `apiToken` below puts an admin token into the querier's process environment, readable from /proc/1/environ, where nothing consumes it.

`webui.enabled` with the session secret is a different matter, that part works and is worth having.

charts/influxdb3-enterprise/templates/explorer-deployment.yaml
```diff
@@ -0,0 +1,148 @@
...
+              cp /config-src/config.json /work-config/config.json
+              {{- end }}
+              chown -R 1500:1500 /db /work-config
+              chmod 700 /db
+              chmod 755 /work-config
+              {{- if get $connection "enabled" }}
+              chmod 644 /work-config/config.json
+              {{- end }}
+          securityContext:
+            runAsUser: 0
```
tomklapka: The init container needs root only for the `chmod` calls, not for the copy. Dropping `chown -R` and the three `chmod` lines and running this as 1500 works, because `fsGroup: 1500` already covers both emptyDirs.

I tested it on the cluster: the pod comes up as `uid=1500(influxui)`, the preconfigured connection still seeds, and with `allowPrivilegeEscalation: false`, `capabilities.drop: ["ALL"]` and `seccompProfile: RuntimeDefault` the Deployment passes a server-side dry run in a namespace enforcing `restricted` Pod Security. Keeping the chmods and running as 1500 fails with `chmod: /db: Operation not permitted`, which is what pointed at them.

For fairness: the rest of the chart does not pass `restricted` today either, so this is not a regression. It is the only part that structurally needs root, though, and it looks avoidable.

charts/influxdb3-enterprise/values.yaml
```diff
@@ -609,6 +609,92 @@ processingEngine:
...
+# preconfigure an Explorer connection to this cluster at startup.
+explorer:
+  enabled: false
+  replicas: 1
+  mode: admin                         # query or admin
+
+  image:
+    registry: docker.io
+    repository: influxdata/influxdb3-ui
+    tag: "1.9.0"
```
tomklapka, line 644: Withdrawing this one - I was wrong. `1.10.1` is the WASM guest that Enterprise downloads from dl.influxdata.com, which is a different artifact from the standalone container image. The published tags for `influxdata/influxdb3-ui` stop at `1.9.0`, with only a `1.10.0-pre` beyond it, so `1.9.0` here is the current release and pinning it is right. I found out by bumping it and watching the pod sit in ImagePullBackOff.

There is a real observation underneath, just not an actionable one: the standalone image trails the guest embedded in the server, so the two Explorer paths are not on the same code. Nothing the chart can do about that.

charts/influxdb3-enterprise/templates/explorer-deployment.yaml
```diff
@@ -0,0 +1,148 @@
...
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.explorer.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      volumes:
+        - name: explorer-db
+          emptyDir: {}
```
tomklapka, line 126: With an emptyDir the Explorer's SQLite goes on every restart. The preconfigured connection re-seeds itself, so the demo path is unaffected, but anything a user saved is gone. A persistence block like the other components have would cover it.

tomklapka (commented):
Nice work, and thanks for splitting the standalone path out - that is the part that actually solves the problem today.

I installed this on a kind cluster rather than only rendering it (v1.31.14, Enterprise 3.11.2, commercial license), and the standalone half works the way you describe. With `networkPolicy.enabled=true` the added ingress rule lets the companion through and a fresh session opens on the preconfigured connection, reporting the cluster as Enterprise 3.11.2. I checked the CNI enforces policies first, an unlabelled pod gets nothing on 8181.

The one substantive objection is `webui.defaultConnection`, which I would drop rather than ship. I traced the mechanism in #830: the guest embedded in 3.11.2 does read the four `DEFAULT_*` values, but `build_wasi_ctx` in the Enterprise host passes only `RESTRICTED_MODE`, `SESSION_SECRET_KEY` and `COOKIE_SECURE`, and there is no `inherit_env`, so neither the mounted config.json nor the pod environment can reach it. On the cluster I confirmed the chart delivers all of it correctly, config.json readable at /app-root/config and all four variables in the container environment, and `/api/server-config` is still `[]`. As it stands the chart would document an option that cannot work on any 3.11.x.

Details on that and three smaller points inline. None of the smaller ones are blockers as far as I am concerned.

One more thing about `webui.enabled` itself, which I missed the first time round. The UI shares port 8181 with the API and is dispatched by path: `should_handle_path` keeps `/api`, `/api/v1|v2|v3`, `/debug`, `/health`, `/metrics`, `/ping`, `/query`, `/ready` and `/write` on the InfluxDB side and hands everything else, `/` included, to the UI. It is a catch-all rather than a list. That branch also sits ahead of the auth layer in `unified_service/service.rs`, so the UI answers unauthenticated. I checked it both ways on a cluster: without `webui` a querier returns 401 for `/` and for `/api/sessions`, and with it enabled the same two paths return the UI and a valid session with no token.

None of that is a defect in this PR, it is how the feature works upstream. It does make enabling `webui` on the queriers a bigger change than it looks, since it opens an unauthenticated surface on a port where everything except `/health` and `/ping` previously needed a token, and it changes what a querier returns for any mistyped path. A dedicated node running `--mode=webui` on its own is supported and would keep that off the query path without scaling the UI against query load. Either shape seems reasonable to me, but the choice deserves a line in the README.

tomklapka:
Correcting my own opening line: "the part that actually solves the problem today" was loose. The companion solves getting a preconfigured Explorer out of Helm, not what #830 reports, which is the integrated UI on the querier ignoring the same configuration. That one needs a host change, so #830 stays open regardless of what happens here.

tomklapka:
We picked this up and finished it, since it had gone quiet and the Explorer work is wanted. Your commit is untouched at the base; everything below is a second commit on top, rebased onto master now that #827 and #828 have landed. Say the word if you would rather take it back.

The one substantive removal is `webui.defaultConnection`. I traced it in #830: Enterprise 3.11.2 passes three variables into the embedded UI and reads no config file, so the mounted `config.json` and the `DEFAULT_*` env never reached it. The option rendered, installed, went green and did nothing, while putting an admin token in the querier's environment. `webui.enabled` with the session secret works and stays.

Two things I found while testing that are worth knowing about the feature itself. The session secret was going onto the command line as `--webui-session-secret=$(ENV)`; kubelet expands that before exec, so it was readable in `/proc/1/cmdline`. Upstream declares the flag with an env var, which the chart already sets, so the flag is gone. And enabling `webui` puts an unauthenticated surface on the querier: the UI takes every path the InfluxDB API does not claim, and that dispatch runs ahead of the auth layer. That is upstream behaviour, not a chart bug, but it is now in the README.

`explorer.mode` now defaults to `query`, matching the Explorer image; `admin` unlocks database and token management, so it is opt-in.

Added: an optional ingress, optional persistence (the SQLite holds saved queries and server configurations and was lost on every restart), a NetworkPolicy for the Explorer, and `explorer.extraVolumes`/`extraEnv` so a private CA can be mounted when the connection is HTTPS.

Fixed along the way: the preconfigured URL hardcoded `http` even with TLS on, the Explorer ran with a mounted Kubernetes API token, a ReadWriteOnce volume with a rolling update deadlocks, `existingClaim` could never work, an explicitly null `explorer` aborted the render, a numeric session secret crashed `b64enc`, and the CI values asked for a trial licence over the internet instead of using the secret the workflow provisions.

The init container your version used to copy `config.json` is gone. The image reads `CONFIG_PATH`, so the Secret mounts read-only at `/etc/influxdb3-explorer` and the variable points at it - no init container, no second emptyDir, no busybox. Worth knowing why the obvious alternative does not work: a `subPath` mount at `/app-root/config/config.json` fails, because the entrypoint refuses to start when that directory exists, is non-empty and is not writable, and in the image it is root-owned.

On your version pin: I withdrew my earlier comment about aligning the image to 1.10.1. That is the WASM guest Enterprise downloads, a different artifact from the container; published tags stop at 1.9.0. You had it right.

Verified on kind v1.31.14 against 3.11.2 with a commercial licence: both UIs serve the real application, create sessions and return query results through their own API; the preconfigured connection seeds; persistence survives a pod restart; and an upgrade from the released 0.11.0 with no Explorer values set changes nothing but the chart label. 37 boundary cases and four rounds of review, which found rather a lot - the diff is bigger than what you opened, so it deserves a careful read.

Chart version is 0.14.0 because it stacks behind #832 and #833; if those do not land first it should come down to 0.12.0.

charts/influxdb3-enterprise/values.yaml
tomklapka: Removed in the takeover commit, along with the secret template, the helpers and the querier env and volume. Verified on a cluster that no DEFAULT_* or CONFIG_PATH is left in the querier.

charts/influxdb3-enterprise/templates/explorer-deployment.yaml
tomklapka: Done. The chown and the chmods are gone, fsGroup covers the volumes, and the init container runs as 1500 with no privilege escalation and no capabilities. It uses the Explorer image now rather than pulling busybox, and the Deployment passes a server-side dry run in a namespace enforcing restricted Pod Security.

charts/influxdb3-enterprise/values.yaml
tomklapka: Left at 1.9.0, as you had it.

charts/influxdb3-enterprise/templates/explorer-deployment.yaml
tomklapka: Added explorer.persistence with an optional PVC, defaulting to the emptyDir as before. Verified that a server configuration survives a pod restart on the volume, and the example now enables it.

tomklapka:
Follow-up on the three Copilot points and one of my own.

`explorer.replicas` now takes only 0 or 1; `foo` used to render as `replicas: foo` and `-1` went through untouched. `explorer.ingress.enabled` without the global `ingress.enabled` used to render nothing and say nothing, and now fails with the reason. The Explorer env goes through the chart's `componentExtraEnv` helper instead of appending both lists, so a name set in both places is emitted once with the component value winning, like every other component.

The NOTES change was mine and was wrong: the Explorer URL block landed inside the ingress-disabled branch, where its own condition could never be true, and it swallowed the write endpoint line. Repaired, and the Explorer line now prints its ingress URL when there is one and the port-forward otherwise.

I also moved the tests into the chart. `tests/explorer_test.yaml` is a helm-unittest suite of twelve cases covering the querier mode, the session secret coming from the environment rather than argv, the quoted-boolean guard, the rootless Explorer with no API token, how it finds its config, persistence, the ingress and its host collision, and the NetworkPolicy pod port. `helm plugin install https://github.com/helm-unittest/helm-unittest` then `helm unittest charts/influxdb3-enterprise`.

Worth considering separately: `kubeconform` over the rendered output catches a class this chart is exposed to. A string in an integer field passes `helm lint` and `helm template` and is only rejected by the API server; that is how `terminationGracePeriodSeconds: abc` got through in #832 until validation caught it. All five ci values files validate clean today. I have not touched the workflow, since that is repo-wide and yours to decide.

Since writing this I ran a simplification pass over the Explorer templates. It removed the init container as described above, dropped `explorer.egressCidrs` - configurability I had added that nobody asked for, and which the chart's other CIDR escape hatches already cover under `networkPolicy.egress.*` - and flattened the validation helpers without changing a single error message. Verified on kind: the pods reach 1/1 and the preconfigured connection still seeds.

tomklapka:
The red CI here is not this branch. It fails because the PR comes from a fork, and it has been failing that way since d316660 on 4 September, before any of my changes.

`helm-charts-test.yaml` triggers on `pull_request` and builds the licence secret from a repository secret:

```
kubectl create secret generic influxdb3-license \
  --from-literal=license-file="${INFLUXDB3_ENTERPRISE_LICENSE_KEY}"
```

GitHub does not expose repository secrets to `pull_request` runs from a fork, so the variable is empty, the Secret is created with an empty value, and every InfluxDB pod logs `failed to verify local jwt: license must be valid jwt: InvalidToken` and never reaches Ready until `ct install` times out. The install fails on the first values file, `auth-existing-secret-values.yaml`, so it never even reaches the Explorer one.

Two things follow. Any chart change that needs the licence cannot get a green run from a fork, whatever the diff contains - #832 and #833 are green today only because they are branches on this repository. And the red mark on this PR carries no information about the code, which is its own problem: nobody reads a check that is always red.

Nothing in the chart can fix it. The options are to move this to a branch on the upstream repo, or to change the workflow - `pull_request_target` with the usual care about untrusted refs, or a fallback to a trial licence when the secret is absent. That second one is repo-wide infrastructure and your call rather than mine.

tomklapka:
Moved this to #834, on a branch in this repository, so CI can actually run - the licence secret is not available to fork PRs, which is why this one has been red since 4 September no matter what the diff contained.

Your commit is the base there unchanged and under your name; my work sits on top as a second commit. Nothing is lost by closing this, and #834 credits the work to you. If you would rather carry it yourself, say so and I will close #834 instead - the branch is `feat/explorer-ui-support` and you are welcome to it either way.

dburton-influxdata:
@tomklapka thank you for all your thorough work. No need for me to take it back. Please run with your branch, PR (834). I assume that GitHub will close out my branch after you commit yours (834).

dburton-influxdata:
Thanks for finishing this — happy to have it land as #834. Appreciate you tracing #830 and fixing the issues you found.

tomklapka:
Closing this. The decision on the product side is webui only: the Web UI built into the server binary rather than the standalone Explorer image, so neither this nor #834, which took it over, will ship as it stands.

Thanks for the groundwork. The ingress and NetworkPolicy parts carry over into the webui PR, which I am starting now against the sources on influxdb_pro main.

Issue #817, opened 2026-07-31
Support InfluxDB 3.11 in Helm Chart


dburton-influxdata:
# Support InfluxDB 3.11 in Helm Chart

## Overview

Update the `influxdb3-enterprise` Helm chart to support InfluxDB 3.11, which introduces breaking changes in environment variables, configuration flags, and default behavior. This is the largest configuration-surface change since InfluxDB 3.0.

**Chart Version:** 0.8.0 → 0.9.0  
**App Version:** 3.9.3 → 3.11.0  
**Release Notes:** [InfluxDB 3.11 Internal Release Notes](https://www.influxdata.com/blog/influxdb-3-11-release/)

---

### 🔴 **Critical - Must Fix for 3.11 Compatibility**

1. **34 Environment Variables Renamed**
   - All `INFLUXDB3_ENTERPRISE_*` → `INFLUXDB3_*`
   - `INFLUXDB3_START_WITHOUT_AUTH` → `INFLUXDB3_WITHOUT_AUTH`
   - `INFLUXDB3_DB_DIR` → `INFLUXDB3_DATA_DIR`
   - `INFLUXDB3_TELEMETRY_DISABLE_UPLOAD` → `INFLUXDB3_DISABLE_TELEMETRY_UPLOAD`
   - Old names work with deprecation warnings, but should be updated

2. **44 PachaTree Configuration Flags Renamed**
   - All `--pt-*` flags lost prefix (e.g., `--pt-shard-count` → `--shard-count`)
   - `--pt-max-columns` → `--max-total-columns`
   - `--pt-gen0-max-bytes-per-file` → `--gen0-max-file-size`
   - `--pt-wal-replica-queue-size` → `--wal-replica-queue-length`

3. **6 Configuration Flags Deleted**
   - `--hard-delete-default-duration` / `INFLUXDB3_HARD_DELETE_DEFAULT_DURATION`
   - `--object-store-cache-endpoint` / `INFLUXDB3_OBJECT_STORE_CACHE_ENDPOINT`
   - `--database-split-level`, `--table-split-level`, `--max-compact-destination`
   - `--pt-enable-row-deletes`

4. **Memory Size Flag Parsing Changed**
   - Bare numbers now **rejected** for 4 flags (to prevent 1024x meaning change)
   - Affected: `execMemPoolBytes`, `fileCacheSize`, `forceSnapshotMemThreshold`, `memoryPoolSize`
[...]

PR #832, opened 2026-09-08
feat(influxdb3-ent): add a termination grace period, document the 3.11 metrics change

tomklapka:
The server drains connections for `shutdown.timeout` while kubelet sends SIGKILL after `terminationGracePeriodSeconds`. Both default to 30 seconds and the chart had no way to set the second one, so a drain that ran its full timeout was cut off at the moment it would have finished.

Adds `shutdown.terminationGracePeriodSeconds` for the ingester, querier, compactor and processor pods. When set, it must be longer than `shutdown.timeout`, or at least the 30s default when no timeout is set. `shutdown.timeout` shipped in 0.10.0, so it is still passed through as written and never fails on its own.

While writing it up I found `telemetry` and `traces` emit their first key unconditionally: setting only `telemetry.endpoint` renders `INFLUXDB3_DISABLE_TELEMETRY_UPLOAD:` with no value and the pods fail to start. That predates this branch, but the new parameter rows made it reachable, so it is guarded here.

Documentation: the `db` label removal in 3.11 now lists the four affected metrics in `UPGRADING-3.10-TO-3.11.md`, where the breaking change was already noted, with the README pointing at it. Seven parameter rows were missing for values the chart already honours (`dataLifecycle`, `logs`, `resourceLimits`, `telemetry`, `traces`).

Verified on kind v1.31.14 with 3.11.2 and a commercial licence: pods reach 1/1, the grace period lands on all of them and the server logs the timeout. `ci/default-values.yaml` now sets both keys, so CI installs a pod carrying them without adding an install run.

Closes part of #817.

### DIFF
--- a/charts/influxdb3-enterprise/templates/_helpers.tpl
+++ b/charts/influxdb3-enterprise/templates/_helpers.tpl
@@ -761,52 +761,58 @@ Permission tokens volumes
 {{- end }}
 
 {{/*
-Termination grace period, shared by every workload.
+Termination grace period for the ingester, querier, compactor and processor pods.
 
-Returns the value or an empty string, so a release that does not set one keeps
-the Kubernetes default of 30s and renders no field at all.
+Returns the value or an empty string. An absent or null key renders no field, so
+the Kubernetes default of 30s still applies.
 */}}
 {{- define "influxdb3-enterprise.terminationGracePeriodValue" -}}
 {{- $shutdown := .Values.shutdown | default dict -}}
 {{- if hasKey $shutdown "terminationGracePeriodSeconds" -}}
-{{- get $shutdown "terminationGracePeriodSeconds" -}}
+{{- $v := get $shutdown "terminationGracePeriodSeconds" -}}
+{{- if not (kindIs "invalid" $v) -}}
+{{- $v -}}
+{{- end -}}
 {{- end -}}
 {{- end }}
 
 {{/*
-Reject a drain that outlives the pod, and a grace period Kubernetes will not take.
+Reject a drain that outlives the pod.
 
 kubelet sends SIGKILL once terminationGracePeriodSeconds expires, so a longer
-shutdown.timeout never finishes draining. Durations built from whole h/m/s parts
-are compared; any other spelling the server accepts is left alone rather than
-guessed at.
+shutdown.timeout never finishes draining. Both sides default to 30, and the
+comparison uses those defaults, so setting either key alone is still checked.
+Durations are built from whole h/m/s parts; the server takes nothing else.
 */}}
 {{- define "influxdb3-enterprise.validateShutdownConfig" -}}
 {{- $shutdown := .Values.shutdown | default dict -}}
-{{- if hasKey $shutdown "terminationGracePeriodSeconds" -}}
-{{- $graceRaw := get $shutdown "terminationGracePeriodSeconds" | toString -}}
+{{- $grace := 30 -}}
+{{- if and (hasKey $shutdown "terminationGracePeriodSeconds") (not (kindIs "invalid" (get $shutdown "terminationGracePeriodSeconds"))) -}}
+{{- $graceVal := get $shutdown "terminationGracePeriodSeconds" -}}
+{{- $graceRaw := $graceVal | toString -}}
 {{- if not (regexMatch "^[0-9]+$" $graceRaw) -}}
 {{- fail (printf "shutdown.terminationGracePeriodSeconds must be a whole number of seconds, got %q. Kubernetes rejects anything else." $graceRaw) -}}
 {{- end -}}
-{{- if hasKey $shutdown "timeout" -}}
-{{- $raw := get $shutdown "timeout" | toString -}}
-{{- $grace := $graceRaw | int64 -}}
-{{- $seconds := -1 -}}
-{{- if regexMatch "^([0-9]+h)?([0-9]+m)?([0-9]+s)?$" $raw -}}
+{{- $grace = $graceRaw | int64 -}}
+{{- end -}}
+{{- $seconds := 30 -}}
+{{- if and (hasKey $shutdown "timeout") (not (kindIs "invalid" (get $shutdown "timeout"))) -}}
+{{- $timeoutVal := get $shutdown "timeout" -}}
+{{- $raw := $timeoutVal | toString -}}
+{{- $parts := regexFindAll "[0-9]+[hms]" $raw -1 -}}
+{{- if or (eq (len $parts) 0) (ne $raw (join "" $parts)) -}}
+{{- fail (printf "shutdown.timeout must be a duration built from whole hour, minute and second parts, such as \"90s\", \"2m\" or \"1m30s\", got %q. The server rejects anything else and the pod will not start." $raw) -}}
+{{- end -}}
 {{- $total := 0 -}}
-{{- range $part := regexFindAll "[0-9]+[hms]" $raw -1 -}}
+{{- range $part := $parts -}}
 {{- $n := trimSuffix "h" (trimSuffix "m" (trimSuffix "s" $part)) | int64 -}}
 {{- if hasSuffix "h" $part -}}{{- $total = add $total (mul $n 3600) -}}
 {{- else if hasSuffix "m" $part -}}{{- $total = add $total (mul $n 60) -}}
 {{- else -}}{{- $total = add $total $n -}}{{- end -}}
 {{- end -}}
-{{- if gt (len (regexFindAll "[0-9]+[hms]" $raw -1)) 0 -}}
 {{- $seconds = $total -}}
 {{- end -}}
-{{- end -}}
-{{- if and (ge $seconds 0) (gt $seconds $grace) -}}
-{{- fail (printf "shutdown.timeout (%s) is longer than shutdown.terminationGracePeriodSeconds (%d): kubelet sends SIGKILL before the drain can finish. Raise the grace period past the timeout, or shorten the timeout." $raw $grace) -}}
-{{- end -}}
-{{- end -}}
+{{- if gt $seconds $grace -}}
+{{- fail (printf "shutdown.timeout resolves to %ds, longer than shutdown.terminationGracePeriodSeconds (%d): kubelet sends SIGKILL before the drain can finish. Both default to 30 when unset. Raise the grace period past the timeout, or shorten the timeout." $seconds $grace) -}}
 {{- end -}}
 {{- end }}
--- a/charts/influxdb3-enterprise/templates/_helpers.tpl
+++ b/charts/influxdb3-enterprise/templates/_helpers.tpl
@@ -786,39 +786,42 @@ Durations are built from whole h/m/s parts; the server takes nothing else.
 */}}
 {{- define "influxdb3-enterprise.validateShutdownConfig" -}}
 {{- $shutdown := .Values.shutdown | default dict -}}
-{{/* Both fall back to 30, the server's drain default and kubelet's. That pairing carries no
-     margin, but it is the product's own default and cannot be a render failure; the margin is
-     required as soon as a release sets either key. */}}
-{{- $explicit := false -}}
+{{/* kubelet's default, which is what the pods get when the chart renders no value. */}}
 {{- $grace := 30 -}}
 {{- if and (hasKey $shutdown "terminationGracePeriodSeconds") (not (kindIs "invalid" (get $shutdown "terminationGracePeriodSeconds"))) -}}
-{{- $explicit = true -}}
-{{- $graceVal := get $shutdown "terminationGracePeriodSeconds" -}}
-{{- $graceRaw := $graceVal | toString -}}
+{{- $graceRaw := get $shutdown "terminationGracePeriodSeconds" | toString -}}
 {{- if not (regexMatch "^[0-9]+$" $graceRaw) -}}
 {{- fail (printf "shutdown.terminationGracePeriodSeconds must be a whole number of seconds, got %q. Kubernetes rejects anything else." $graceRaw) -}}
 {{- end -}}
 {{- $grace = $graceRaw | int64 -}}
 {{- end -}}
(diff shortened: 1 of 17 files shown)


charts/influxdb3-enterprise/templates/_helpers.tpl
```diff
@@ -759,3 +759,60 @@ Permission tokens volumes
...
+{{- $grace = $graceRaw | int64 -}}
+{{- end -}}
+{{- $seconds := 30 -}}
+{{- if and (hasKey $shutdown "timeout") (not (kindIs "invalid" (get $shutdown "timeout"))) -}}
+{{- $timeoutVal := get $shutdown "timeout" -}}
+{{- $raw := $timeoutVal | toString -}}
+{{- $parts := regexFindAll "[0-9]+[hms]" $raw -1 -}}
+{{- if or (eq (len $parts) 0) (ne $raw (join "" $parts)) -}}
+{{- fail (printf "shutdown.timeout must be a duration built from whole hour, minute and second parts, such as \"90s\", \"2m\" or \"1m30s\", got %q. The server rejects anything else and the pod will not start." $raw) -}}
+{{- end -}}
```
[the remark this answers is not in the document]
tomklapka: Good catch, fixed. A null now renders nothing rather than an empty env var, which matches how the grace period already handled it - setting a key to null is the documented way to clear an inherited value on upgrade, so failing would be the wrong call. Tests cover both the null timeout and a null alongside a set grace period. 1c947fc

charts/influxdb3-enterprise/templates/configmap.yaml
```diff
@@ -294,8 +297,10 @@ data:
 
   # Telemetry
   {{- with .Values.telemetry }}
+  {{- if hasKey . "disableUpload" }}
```
bednar: `hasKey` is also true when `disableUpload` is explicitly `null`, so this renders both telemetry variables with an empty value. A layered values file can produce this when clearing an inherited key, and the server may then reject the empty boolean configuration. Please keep the current handling of explicit `false`, but skip the variables when the value is null (for example, combine `hasKey` with a `kindIs "invalid"` check).

charts/influxdb3-enterprise/templates/_helpers.tpl
```diff
@@ -759,3 +759,60 @@ Permission tokens volumes
...
+{{- $total := 0 -}}
+{{- range $part := $parts -}}
+{{- $n := trimSuffix "h" (trimSuffix "m" (trimSuffix "s" $part)) | int64 -}}
+{{- if hasSuffix "h" $part -}}{{- $total = add $total (mul $n 3600) -}}
+{{- else if hasSuffix "m" $part -}}{{- $total = add $total (mul $n 60) -}}
+{{- else -}}{{- $total = add $total $n -}}{{- end -}}
+{{- end -}}
+{{- $seconds = $total -}}
+{{- end -}}
+{{- if gt $seconds $grace -}}
```
bednar: Could this reject equality as well (or document equality as safe)? The chart documentation says the grace period has to outlast the drain, but `gt` allows `shutdown.timeout: 60s` with `terminationGracePeriodSeconds: 60`. With both deadlines equal there is no margin for the process to exit before kubelet's termination deadline, so the drain can still be cut off. Please require the grace period to be strictly greater than the timeout, consistent with the documentation.

bednar (changes requested):
Please address the two inline review comments before merging.

charts/influxdb3-enterprise/templates/configmap.yaml
tomklapka: Fixed, hasKey paired with a kindIs "invalid" check. Clearing the key with null renders neither variable; an explicit false still renders both. 72bda61

charts/influxdb3-enterprise/templates/_helpers.tpl
tomklapka: Done, but not unconditionally - both defaults are 30, so a blanket rule would reject every stock install. The check now runs only when a release sets shutdown.timeout, and a zero timeout skips it since the server treats that as "no drain". Also widened the format: shutdown_timeout is humantime, and the old regex rejected 500ms, "1h 30m" and "2 minutes". be8b8b9

One thing to confirm: I still reject a bare 0, since I could not verify humantime takes it. Say the word and it goes.

charts/influxdb3-enterprise/templates/_helpers.tpl
```diff
@@ -759,3 +759,69 @@ Permission tokens volumes
...
+{{- if and (hasKey $shutdown "terminationGracePeriodSeconds") (not (kindIs "invalid" (get $shutdown "terminationGracePeriodSeconds"))) -}}
+{{- $graceRaw := get $shutdown "terminationGracePeriodSeconds" | toString -}}
+{{- if not (regexMatch "^[0-9]+$" $graceRaw) -}}
+{{- fail (printf "shutdown.terminationGracePeriodSeconds must be a whole number of seconds, got %q. Kubernetes rejects anything else." $graceRaw) -}}
+{{- end -}}
+{{- $grace = $graceRaw | int64 -}}
+{{- end -}}
+{{/* Only a timeout the release sets can be compared. The server picks its own default when the
+     key is absent, and writing a default down must not change the verdict. */}}
+{{- if and (hasKey $shutdown "timeout") (not (kindIs "invalid" (get $shutdown "timeout"))) -}}
```
bednar, line 802: When `shutdown.timeout` is absent, this entire comparison is skipped even when the chart renders an explicit grace period. For example, `shutdown.terminationGracePeriodSeconds: 10` renders successfully, but the server still uses its default `shutdown_timeout=30s`; kubelet can therefore SIGKILL the pod during the drain. Please compare an explicit grace period against the effective 30s server default too, while keeping both values unset valid.

charts/influxdb3-enterprise/templates/_helpers.tpl
```diff
@@ -759,3 +759,69 @@ Permission tokens volumes
...
+{{- $raw := get $shutdown "timeout" | toString | trim | lower -}}
```
bednar: Validation lowercases the value, but `configmap.yaml` emits the original value. This changes the meaning of case-sensitive humantime units: `shutdown.timeout: "1M"` is checked as one minute but passed to the server as `1M` (one month). With a grace period such as 120s this passes Helm validation and still gets killed during the actual drain. Please validate the exact string that is rendered, without changing unit semantics.

charts/influxdb3-enterprise/templates/_helpers.tpl
```diff
@@ -759,3 +759,69 @@ Permission tokens volumes
...
+{{/* shutdown_timeout is a humantime::Duration, so match the units humantime takes. */}}
+{{- $unit := "nsec|ns|usec|us|msec|ms|seconds|second|secs|sec|s|minutes|minute|mins|min|m|hours|hour|hrs|hr|h|days|day|d" -}}
+{{- if not (regexMatch (printf "^([0-9]+\\s*(%s)\\s*)+$" $unit) $raw) -}}
```
bednar: The allow-list is narrower than the server's `humantime::Duration` grammar. The server accepts e.g. `1week`, `1month`, `1year`, `1nanos`, and `1millis`, but this regex rejects them at render time. Please keep the chart validation aligned with the parser actually used by `shutdown_timeout`, otherwise valid server configurations cannot be deployed.

charts/influxdb3-enterprise/templates/_helpers.tpl
```diff
@@ -759,3 +759,69 @@ Permission tokens volumes
...
+{{- fail (printf "shutdown.timeout must be a duration the server can parse, such as \"90s\", \"2m\", \"1m30s\" or \"500ms\", got %q." $raw) -}}
+{{- end -}}
+{{- $ms := 0 -}}
+{{- range $part := regexFindAll (printf "[0-9]+\\s*(%s)" $unit) $raw -1 -}}
+{{- $n := regexFind "[0-9]+" $part | int64 -}}
+{{- $u := regexReplaceAll "[0-9]+\\s*" $part "" -}}
+{{- if or (eq $u "ms") (eq $u "msec") -}}{{- $ms = add $ms $n -}}
+{{- else if or (hasPrefix "n" $u) (hasPrefix "u" $u) -}}
```
bednar: These branches discard any non-zero `nsec`/`ns`/`usec`/`us` value from the comparison. In particular, `shutdown.timeout: 2us` with `terminationGracePeriodSeconds: 0` passes validation and is rendered, even though the server still has a non-zero drain. Please preserve sub-millisecond precision (or at least treat any positive duration as greater than zero).

charts/influxdb3-enterprise/templates/_helpers.tpl
```diff
@@ -759,3 +759,69 @@ Permission tokens volumes
...
+{{- if and (hasKey $shutdown "timeout") (not (kindIs "invalid" (get $shutdown "timeout"))) -}}
+{{- $raw := get $shutdown "timeout" | toString | trim | lower -}}
+{{/* shutdown_timeout is a humantime::Duration, so match the units humantime takes. */}}
+{{- $unit := "nsec|ns|usec|us|msec|ms|seconds|second|secs|sec|s|minutes|minute|mins|min|m|hours|hour|hrs|hr|h|days|day|d" -}}
+{{- if not (regexMatch (printf "^([0-9]+\\s*(%s)\\s*)+$" $unit) $raw) -}}
+{{- fail (printf "shutdown.timeout must be a duration the server can parse, such as \"90s\", \"2m\", \"1m30s\" or \"500ms\", got %q." $raw) -}}
+{{- end -}}
+{{- $ms := 0 -}}
+{{- range $part := regexFindAll (printf "[0-9]+\\s*(%s)" $unit) $raw -1 -}}
+{{- $n := regexFind "[0-9]+" $part | int64 -}}
```
bednar: `int64` conversion can overflow before the comparison. For example, a timeout such as `999999999999999999999999999999999999s` renders successfully because `$n` becomes zero after conversion, while the original oversized value is still emitted to the ConfigMap. This should fail like humantime's number-overflow error; the same range check is needed for the grace-period conversion at line 796.

bednar (changes requested):
Please address all five inline review comments before merging.

charts/influxdb3-enterprise/templates/_helpers.tpl
tomklapka: Done. A grace period without a timeout is compared with the 30s default, so 10 fails and 30 passes. 72b3f1b

charts/influxdb3-enterprise/templates/_helpers.tpl
tomklapka: The timeout is no longer lowercased, so 1M reads as a month. 72b3f1b

charts/influxdb3-enterprise/templates/_helpers.tpl
tomklapka: shutdown.timeout has shipped since 0.10.0, so it no longer fails on its own; this branch had broken upgrades for the chart's own "30s" example. The comparison now uses humantime's full unit table. 3641056, 72b3f1b

charts/influxdb3-enterprise/templates/_helpers.tpl
tomklapka: Sub-second parts round up to a millisecond, so 2us against a grace period of 0 fails. 72b3f1b

charts/influxdb3-enterprise/templates/_helpers.tpl
tomklapka: Timeout parts are capped at six digits and read with atoi, and the grace period at 999999999. 72b3f1b

charts/influxdb3-enterprise/templates/_helpers.tpl
```diff
@@ -759,3 +759,74 @@ Permission tokens volumes
...
+{{/* kubelet holds the grace period as nanoseconds in an int64, which ends near 9.2e9 seconds. */}}
+{{- if not (regexMatch "^(0|[1-9][0-9]{0,8})$" $grace) -}}
+{{- fail (printf "shutdown.terminationGracePeriodSeconds must be a whole number of seconds from 0 to 999999999, got %q." $grace) -}}
+{{- end -}}
+{{- $graceMs := mul ($grace | atoi) 1000 -}}
+{{- if and (hasKey $shutdown "timeout") (not (kindIs "invalid" (get $shutdown "timeout"))) -}}
+{{- $raw := get $shutdown "timeout" | toString | trim -}}
+{{/* humantime 2.3.0's units, longest spelling first so regexFindAll does not stop at m in ms. */}}
+{{- $unit := "nanos|nsec|ns|usec|us|µs|millis|msec|ms|seconds|second|secs|sec|s|minutes|minute|mins|min|m|hours|hour|hrs|hr|h|days|day|d|weeks|week|wks|wk|w|months|month|M|years|year|yrs|yr|y" -}}
+{{- if regexMatch (printf "^([0-9]{1,6} *(%s) *)+$" $unit) $raw -}}
```
bednar: [P1] Do not silently skip valid humantime values here. The six-digit/integer-only regex excludes supported values such as `1.5h` and `1000000s`; with `terminationGracePeriodSeconds: 10`, Helm accepts and renders either value even though the server can drain for longer than the grace period, allowing kubelet to send SIGKILL mid-drain. Please compare the complete supported duration grammar with overflow-safe arithmetic, or reject values that cannot be compared whenever an explicit grace period is set, and add regression coverage.

bednar (changes requested):
Please address the inline comment before merging.

charts/influxdb3-enterprise/templates/_helpers.tpl
tomklapka: With a grace period set, a timeout the chart cannot read exactly now fails instead of being skipped, so 1.5h is rejected. Parts up to 18 digits are compared, capped above the largest allowed grace period, so 1000000s is compared without overflowing. Without a grace period the timeout is still passed through. 08a2b22

bednar (approved):
Approved.

Issue #804, opened 2026-06-24
Improve memory configuration documentation and validation for InfluxDB3 Enterprise


dburton-influxdata:
---
title: "Improve memory configuration documentation and validation for InfluxDB3 Enterprise"
labels:
  - documentation
  - enhancement
  - influxdb3-enterprise
---

## Summary

Add comprehensive memory configuration documentation and validation to prevent common OOM (Out of Memory) kills in InfluxDB 3 Enterprise deployments, particularly affecting the compactor component.

## Problem

Customer deployments are experiencing OOM kills due to memory overcommitment from misconfigured cache settings. Current issues:

1. **No memory configuration guidance** in HELM chart documentation
2. **No validation** of memory settings during chart installation
3. **Percentage-based cache configuration is ambiguous** (`INFLUXDB3_PT_FILE_CACHE_MAX_BYTES: 50%`)
4. **Conflicting defaults** that lead to cache allocation > container limit
5. **No warnings** when force snapshot threshold is misconfigured

### Real-World Example (VAST Customer)

Customer configured:
```yaml
resources:
  limits:
    memory: 95Gi

env:
  - name: INFLUXDB3_PT_FILE_CACHE_MAX_BYTES
    value: 50%                              # 47.5GB
  - name: INFLUXDB3_OBJECT_STORE_CACHE_MAX_SIZE_BYTES
    value: "34359738368"                    # 32GB
  - name: INFLUXDB3_PT_COMPACTOR_INPUT_SIZE_BUDGET
    value: 12GB
```

**Result:**
- Total cache allocation: 91.5GB (96% of limit)
- Available for operations: 3.5GB (4% of limit)
- **Compactor OOMKilled after 38 minutes**

**Root cause:** Caches alone consumed 96% of available memory, leaving insufficient space for compaction operations.

### 1. Add Memory Sizing Documentation

Create comprehensive memory sizing guide covering:
- Memory budget calculation
- Component-by-component sizing guidelines
- Pre-configured profiles (small/standard/large/high-performance)
- Troubleshooting guide
- Validation checklist

**Documentation structure:**
```
docs/
  memory-sizing-guide.md          # Comprehensive guide with calculator
  memory-troubleshooting.md       # OOM troubleshooting
  examples/
    memory-config-64gb.yaml       # Small deployment
    memory-config-95gb.yaml       # Standard deployment
    memory-config-150gb.yaml      # Large deployment
    memory-config-200gb.yaml      # High-performance deployment
```

[...]

PR #833, opened 2026-09-08
feat(influxdb3-ent): reject memory sizes the server will not parse

tomklapka:
InfluxDB takes `b`, `kb`, `mb`, `gb` and `tb`, all 1024-based, or a whole percentage. Bare numbers have been rejected since 3.11, where they used to mean megabytes, and Kubernetes suffixes such as `Gi` are not in the server's unit table at all. Both reached the container and crash-looped it; they now fail the render with the reason. The deprecated aliases keep their old lenient parsing, so a values file carried over from 3.10 still installs.

This covers the "fail fast if bare numbers detected in memory flags" item from #817 Phase 1.

## What this deliberately does not do

#804 asks for a memory budget check, and I built one before finding that the server already has this and scopes it much more narrowly. In `pacha_tree.rs` it sums reservations only for query-only nodes, with the comment that ingest and compaction "share buffers in ways that aren't summable as cleanly". It counts `exec-mem-pool-size`, `file-cache-size` and `replica-max-buffer-size` - not the snapshot threshold - and `replica-max-buffer-size` defaults to half the limit, capped at 16 GiB. And it warns rather than refusing to start.

A chart-side sum with a different formula would have refused valid releases on ingesters and compactors while giving a false all-clear on the queriers the server actually checks, and as a hard `fail` it would have blocked `helm upgrade` for anyone deliberately overcommitting today. The README now documents what the server does, including the warning to grep for, which seemed more useful to the customers in #804 than an invented model.

Happy to revisit if you would rather have the check with the server's exact formula and scope.

Verified on kind v1.31.14 with 3.11.2 and a commercial licence: pods reach 1/1 and the server parses and logs the values the chart accepted. 19 boundary cases cover the accepted and rejected forms, the aliases, unset, null and `0`.

Stacks behind #832, which takes 0.12.0; this is 0.13.0.

Refs #804, #817.

### DIFF
--- a/charts/influxdb3-enterprise/templates/_helpers.tpl
+++ b/charts/influxdb3-enterprise/templates/_helpers.tpl
@@ -759,3 +759,92 @@ Permission tokens volumes
         path: permission-tokens.json
 {{- end }}
 {{- end }}
+
+{{/*
+Convert a size value to bytes.
+
+Accepts InfluxDB size units (b, kb, mb, gb, tb - binary multiples), Kubernetes
+quantities (Ki, Mi, Gi, Ti and K, M, G, T), a bare number read as bytes, or a
+percentage of "ref".
+
+Input: dict with "value" and "ref"; "ref" is in bytes and only percentages need it.
+Returns bytes, or an empty string when the value cannot be parsed.
+*/}}
+{{- define "influxdb3-enterprise.toBytes" -}}
+{{- $v := .value | toString | trim | lower -}}
+{{- $ref := .ref | default 0 | float64 -}}
+{{- if eq $v "" -}}
+{{- else if hasSuffix "%" $v -}}
+{{- $pct := trimSuffix "%" $v -}}
+{{- if and (regexMatch "^[0-9]+(\\.[0-9]+)?$" $pct) (gt $ref 0.0) -}}
+{{- divf (mulf (float64 $pct) $ref) 100.0 | int64 -}}
+{{- end -}}
+{{- else -}}
+{{- $num := regexFind "^[0-9]+(\\.[0-9]+)?" $v -}}
+{{- $unit := regexFind "[a-z]*$" $v -}}
+{{- $mult := dict "" 1 "b" 1 "kb" 1024 "mb" 1048576 "gb" 1073741824 "tb" 1099511627776 "ki" 1024 "mi" 1048576 "gi" 1073741824 "ti" 1099511627776 "k" 1000 "m" 1000000 "g" 1000000000 "t" 1000000000000 -}}
+{{- if and $num (hasKey $mult $unit) -}}
+{{- mulf (float64 $num) (float64 (get $mult $unit)) | int64 -}}
+{{- end -}}
+{{- end -}}
+{{- end }}
+
+{{/*
+Human-readable byte count, for error messages.
+*/}}
+{{- define "influxdb3-enterprise.humanBytes" -}}
+{{- $b := . | float64 -}}
+{{- if ge $b 1073741824.0 -}}
+{{- printf "%.1fGB" (divf $b 1073741824.0) -}}
+{{- else if ge $b 1048576.0 -}}
+{{- printf "%.1fMB" (divf $b 1048576.0) -}}
+{{- else -}}
+{{- printf "%dB" ($b | int64) -}}
+{{- end -}}
+{{- end }}
+
+{{/*
+Reject a memory size the server will not take.
+
+InfluxDB accepts b, kb, mb, gb and tb - all binary, 1024-based - or a whole
+percentage, with optional whitespace before the suffix. Bare numbers were
+rejected in 3.11 because they used to mean megabytes, and Kubernetes suffixes
+such as Gi are not in the server's table at all, so both reach the container and
+crash-loop it. The deprecated aliases keep the old lenient parsing and are not
+checked.
+
+Whether the configured sizes add up to more than the container can hold is the
+server's call, not the chart's: it sums them only for query-only nodes, using
+its own set of reservations, and warns rather than refusing to start. See the
+Memory Budget section of the README.
+*/}}
+{{- define "influxdb3-enterprise.validateMemorySize" -}}
+{{- $key := .key -}}
+{{- $raw := .value | toString | trim | lower -}}
+{{- if ne $raw "" -}}
+{{- if hasSuffix "%" $raw -}}
+{{- $pct := trimSuffix "%" $raw | trim -}}
+{{- if not (regexMatch "^[0-9]+$" $pct) -}}
+{{- fail (printf "%s must be a whole percentage such as \"20%%\", got %q." $key .value) -}}
+{{- end -}}
+{{- if gt ($pct | int64) 100 -}}
+{{- fail (printf "%s must be between 0 and 100 percent, got %q." $key .value) -}}
+{{- end -}}
+{{- else if not (regexMatch "^[0-9]+ ?(kb|mb|gb|tb|b)$" $raw) -}}
+{{- fail (printf "%s must carry a unit suffix the server accepts (b, kb, mb, gb, tb - all 1024-based) or be a percentage, got %q. Bare numbers and Kubernetes suffixes such as Gi are rejected by InfluxDB and the pod will not start." $key .value) -}}
+{{- end -}}
+{{- end -}}
+{{- end }}
+
+{{/*
+Check every memory size a release sets, whatever its limits look like.
+*/}}
+{{- define "influxdb3-enterprise.validateMemorySizes" -}}
+{{- $caching := .Values.caching | default dict -}}
+{{- include "influxdb3-enterprise.validateMemorySize" (dict "key" "caching.fileCacheSize" "value" (get $caching "fileCacheSize" | default "")) -}}
+{{- range $name := list "ingester" "querier" "compactor" "processingEngine" -}}
+{{- $memory := get (get $.Values $name | default dict) "memory" | default dict -}}
+{{- include "influxdb3-enterprise.validateMemorySize" (dict "key" (printf "%s.memory.execMemPoolSize" $name) "value" (get $memory "execMemPoolSize" | default "")) -}}
+{{- include "influxdb3-enterprise.validateMemorySize" (dict "key" (printf "%s.memory.forceSnapshotMemSize" $name) "value" (get $memory "forceSnapshotMemSize" | default "")) -}}
+{{- end -}}
+{{- end }}
--- a/charts/influxdb3-enterprise/templates/_helpers.tpl
+++ b/charts/influxdb3-enterprise/templates/_helpers.tpl
@@ -780,42 +780,75 @@ Memory Budget section of the README.
 {{- define "influxdb3-enterprise.validateMemorySize" -}}
 {{- $key := .key -}}
 {{- $raw := .value | toString | trim | lower -}}
-{{/* Report the value as the user wrote it, and as a string: %q on a YAML number renders as %!q(float64=...). */}}
+{{/* Quote the value as the user wrote it. toString on a values-file number goes through %v,
+     which switches to exponent form above six digits, so integral floats are printed whole. */}}
 {{- $shown := .value | toString -}}
-{{- $allowPercent := ne (.allowPercent | toString) "false" -}}
-{{- if ne $raw "" -}}
+{{- if and (kindIs "float64" .value) (eq (floor .value) .value) -}}
+{{- $shown = printf "%.0f" .value -}}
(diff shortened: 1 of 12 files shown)


charts/influxdb3-enterprise/templates/_helpers.tpl
```diff
@@ -759,3 +759,92 @@ Permission tokens volumes
...
+*/}}
+{{- define "influxdb3-enterprise.validateMemorySizes" -}}
+{{- $caching := .Values.caching | default dict -}}
+{{- include "influxdb3-enterprise.validateMemorySize" (dict "key" "caching.fileCacheSize" "value" (get $caching "fileCacheSize" | default "")) -}}
+{{- range $name := list "ingester" "querier" "compactor" "processingEngine" -}}
+{{- $memory := get (get $.Values $name | default dict) "memory" | default dict -}}
+{{- include "influxdb3-enterprise.validateMemorySize" (dict "key" (printf "%s.memory.execMemPoolSize" $name) "value" (get $memory "execMemPoolSize" | default "")) -}}
+{{- include "influxdb3-enterprise.validateMemorySize" (dict "key" (printf "%s.memory.forceSnapshotMemSize" $name) "value" (get $memory "forceSnapshotMemSize" | default "")) -}}
+{{- end -}}
+{{- end }}
```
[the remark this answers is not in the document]
tomklapka: Right, and wider than one key. The chart renders eight options the server parses as MemorySize and the check reached three of them, so replicaMaxBufferSize, walBufferSize, snapshotSize, mergeThresholdSize and compactorInputSizeBudget are covered now. adb219e

charts/influxdb3-enterprise/templates/_helpers.tpl
```diff
@@ -759,3 +759,92 @@ Permission tokens volumes
         path: permission-tokens.json
 {{- end }}
 {{- end }}
+
+{{/*
+Convert a size value to bytes.
+
+Accepts InfluxDB size units (b, kb, mb, gb, tb - binary multiples), Kubernetes
+quantities (Ki, Mi, Gi, Ti and K, M, G, T), a bare number read as bytes, or a
```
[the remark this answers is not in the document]
tomklapka: Both removed. They were the parser for the budget check that this PR dropped and nothing else used them. adb219e

charts/influxdb3-enterprise/values.yaml
```diff
@@ -249,7 +249,9 @@ ingester:
     # - name: INFLUXDB3_UNSET_VARS
     #   value: "INFLUXDB3_PLUGIN_DIR"
 
-  # Memory tuning
+  # Memory tuning. Sizes take b, kb, mb, gb or tb - all 1024-based - or a whole
+  # percentage of the memory limit. Bare numbers and Kubernetes suffixes such as
+  # Gi are rejected by the server, so the chart fails the render on them.
   # memory:
   #   execMemPoolSize: "20%"              # Deprecated alias: execMemPoolBytes
   #   forceSnapshotMemSize: "50%"         # Deprecated alias: forceSnapshotMemThreshold
```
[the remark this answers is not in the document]
tomklapka: Reworded. The comment now says the listed keys reject bare numbers and that their deprecated aliases keep the older, more permissive parsing, rather than implying it applies everywhere. adb219e

charts/influxdb3-enterprise/templates/_helpers.tpl
```diff
@@ -759,3 +759,55 @@ Permission tokens volumes
...
+{{- define "influxdb3-enterprise.validateMemorySizes" -}}
+{{- $caching := .Values.caching | default dict -}}
+{{- include "influxdb3-enterprise.validateMemorySize" (dict "key" "caching.fileCacheSize" "value" (get $caching "fileCacheSize" | default "")) -}}
+{{- range $name := list "ingester" "querier" "compactor" "processingEngine" -}}
+{{- $memory := get (get $.Values $name | default dict) "memory" | default dict -}}
+{{- include "influxdb3-enterprise.validateMemorySize" (dict "key" (printf "%s.memory.execMemPoolSize" $name) "value" (get $memory "execMemPoolSize" | default "")) -}}
+{{- include "influxdb3-enterprise.validateMemorySize" (dict "key" (printf "%s.memory.forceSnapshotMemSize" $name) "value" (get $memory "forceSnapshotMemSize" | default "")) -}}
+{{- end -}}
+{{- $pacha := get (get .Values "engine" | default dict) "pachaTree" | default dict -}}
+{{- range $k := list "replicaMaxBufferSize" "walBufferSize" "snapshotSize" "mergeThresholdSize" "compactorInputSizeBudget" -}}
```
bednar: [P1] `validateMemorySizes` still misses size options that this chart renders. `pachaTreeEnv` emits `gen0MaxFileSize` and the L1-L4 `*TailTargetSize`/`*TargetFileSize` values, but none of them are included in this validation list. In InfluxDB 3.11 these size options require an explicit unit, so for example `engine.pachaTree.l1TargetFileSize: 4096` passes Helm rendering and then makes the pod fail during startup. Please add these fields (or validate the complete set of rendered size options).

charts/influxdb3-enterprise/templates/_helpers.tpl
```diff
@@ -759,3 +759,55 @@ Permission tokens volumes
...
+{{- if ne $raw "" -}}
+{{- if hasSuffix "%" $raw -}}
+{{- $pct := trimSuffix "%" $raw | trim -}}
+{{- if not (regexMatch "^[0-9]+$" $pct) -}}
+{{- fail (printf "%s must be a whole percentage such as \"20%%\", got %q." $key .value) -}}
+{{- end -}}
+{{- if gt ($pct | int64) 100 -}}
+{{- fail (printf "%s must be between 0 and 100 percent, got %q." $key .value) -}}
+{{- end -}}
+{{- else if not (regexMatch "^[0-9]+ ?(kb|mb|gb|tb|b)$" $raw) -}}
```
bednar: [P2] The server parser allows arbitrary whitespace before the suffix (for example `5  mb`), but this regex accepts at most one literal space. That makes the chart reject values that InfluxDB 3.11 accepts, despite the helper documentation saying whitespace is allowed. Please use a whitespace expression that matches the server parser, such as zero or more whitespace characters.

bednar (changes requested):
Request changes: please address the two inline findings above. The current validation misses several rendered size options, and the regex rejects whitespace forms accepted by InfluxDB 3.11.

charts/influxdb3-enterprise/templates/_helpers.tpl
tomklapka: All 13 MemorySize options are covered now, not just the 5. gen0MaxFileSize is ByteSize, so it takes a unit but no percentage and has its own check. 70ed1bb

Writing the tests turned up a second hole: `get ... | default ""` skipped 0 and false, so l1TargetFileSize: 0 still reached the pod. 55fd89f

charts/influxdb3-enterprise/templates/_helpers.tpl
tomklapka: Fixed, and it is looser than one space. parse_unit_suffix strips the suffix and calls num_str.trim(), so the server takes any run of whitespace and a tab too. The regex is \s* now, and a leading + is accepted since usize::from_str takes one. 70ed1bb

charts/influxdb3-enterprise/templates/_helpers.tpl
```diff
@@ -759,3 +759,96 @@ Permission tokens volumes
...
+{{- end -}}
+{{- if hasSuffix "%" $raw -}}
+{{- if not $allowPercent -}}
+{{- fail (printf "%s is an absolute file size and does not take a percentage, got %q. Use %s." $key $shown $units) -}}
+{{- end -}}
+{{- $pct := trimSuffix "%" $raw | trim | trimPrefix "+" -}}
+{{- if not (regexMatch "^[0-9]+$" $pct) -}}
+{{- fail (printf "%s must be a whole percentage such as \"20%%\", got %q." $key $shown) -}}
+{{- end -}}
+{{- if gt ($pct | int64) 100 -}}
```
bednar: [P1] Guard the percentage range check against integer overflow

Converting an arbitrary digit string with `int64` makes values above Int64 max wrap to 0 (for example, `999999999999999999999999999999999999999999%` currently renders successfully). That bypasses the new 0–100 validation and sends an invalid percentage to InfluxDB, so the pod can still crash-loop. Please validate the decimal string without a lossy integer conversion (or reject values longer than three digits before parsing) and add a regression test.

charts/influxdb3-enterprise/templates/_helpers.tpl
```diff
@@ -759,3 +759,96 @@ Permission tokens volumes
...
+Check every memory size a release sets, whatever its limits look like.
+
+Only keys the release actually sets are checked, and a key cleared with null is
+left alone. Gating on truthiness instead would skip 0 and false, which are the
+values that reach the pod and stop it starting.
+*/}}
+{{- define "influxdb3-enterprise.validateMemorySizeKey" -}}
+{{- $src := .src -}}
+{{- $name := .name -}}
+{{- if and (hasKey $src $name) (not (kindIs "invalid" (get $src $name))) -}}
```
bednar, line 834: [P1] Do not render null PachaTree values

This helper skips an explicit YAML `null`, but `pachaTreeEnv` checks only `hasKey` and consequently emits a null value as `value:`. Thus `--set engine.pachaTree.l1TargetFileSize=null` passes Helm rendering and produces an empty environment variable instead of treating the key as unset. Please skip null values consistently in `pachaTreeEnv` (or otherwise normalize this before rendering) and add a regression test.

charts/influxdb3-enterprise/values.yaml
```diff
@@ -177,6 +177,10 @@ license:
 # - ingester.wal.flushInterval
 # - caching.fileCacheSize, parquetMemCacheQueryPathDuration,
 #   and disableParquetMemCache
+#
+# The *Size and *Budget keys below take b, kb, mb, gb or tb - all 1024-based -
```
bednar, line 181: [P2] Correct the fileCacheSize deprecation comment

This new documentation correctly treats the new `*Size` keys as preferred, but the caching example later in this file (around line 683) still says `fileCacheSize` is the deprecated alias. In 3.11 `fileCacheSize` is the preferred key and `parquetMemCacheSize` is the legacy alias; please update that comment so the values file does not tell users to migrate away from the supported key.

bednar (changes requested):
Please address the three inline findings below before merging. The CI-test execution finding was intentionally omitted per request.

charts/influxdb3-enterprise/templates/_helpers.tpl
tomklapka: Fixed. Sprig int64 turned out not to wrap: it returns 0 on overflow and reads a leading zero as octal, so 0120% also passed as 80. The percentage is now checked as text with leading zeros dropped and bounded before converting, with tests for both. 4849b71

charts/influxdb3-enterprise/templates/_helpers.tpl
tomklapka: Fixed, pachaTreeEnv skips a null, so a cleared key is unset. Test added. 4849b71

charts/influxdb3-enterprise/values.yaml
tomklapka: The comment named parquetMemCacheSize as the deprecated alias, same format as the execMemPoolSize lines, but it reads both ways. They all say "Replaces deprecated X" now. 4849b71

bednar (changes requested):
The new helm-unittest cases are not executed by CI. The workflow currently runs only ct lint and ct install, so the 22 added tests can regress without being detected. Please install/configure helm-unittest in the PR workflow and run the chart test suite there.

tomklapka:
The workflow change covers every chart, so it is in its own PR: #845. It runs the suites from this PR, #832 and #834 once they land.

bednar (dismissed):
Approved. Note: GitHub currently reports this PR as conflicting with master (DIRTY), so the merge conflict still needs to be resolved before it can be merged.

tomklapka:
Merged master to resolve the conflict with #832. Both sides were additive, so the merge keeps both validators, both README sections and both helper groups; the version stays 0.13.0. Nothing else changed since your approval, but the push dismissed it. 6626a8f

bednar (dismissed):

tomklapka:
Merged master to resolve the `version:` conflict after #847 released 0.12.1. Kept 0.13.0, which is still a valid increment; nothing else changed. Sorry for the dismissed approval - the push does that even for a merge commit.

bednar (approved):

PR #834, opened 2026-09-09
feat(influxdb3-ent): add Explorer UI support

tomklapka:
Explorer UI support for the Enterprise chart, continuing @dburton-influxdata's work in #831. His commit is the base here unchanged; the second commit finishes it. This lives on a branch in this repository because CI cannot pass from a fork - the licence secret is not exposed to `pull_request` runs from forks, so every InfluxDB pod fails licence verification and `ct install` times out. That is why #831 has been red since 4 September regardless of its contents.

The chart runs the standalone Explorer as a companion Deployment with its own Service, and optionally an Ingress and a PVC. It does not enable the Web UI embedded in the server: this chart always splits the roles, so that UI would sit on a query-only node where it cannot write, cannot be preconfigured on 3.11.2 (#830), and answers ahead of the auth layer. The comment below has the detail.

`explorer.mode` defaults to `query`, matching the image; `admin` unlocks database and token management and is opt-in. `explorer.defaultConnection.server` defaults to `ingress.host` when the ingress is enabled, because the ingress is the one endpoint that routes both writes and queries; set it explicitly for any other routing.

The standalone Explorer runs rootless with no Kubernetes API token, and its config arrives through `CONFIG_PATH` pointing at a read-only Secret mount, so there is no init container and no busybox dependency. Persistence is optional and off by default; without it the SQLite holding saved queries and server configurations is lost on every restart.

Verified on kind v1.31.14 against 3.11.2 with a commercial licence: both UIs serve the application, create sessions and return query results through their own API; the preconfigured connection seeds; state survives a pod restart on the PVC; the Explorer reaches the querier under its NetworkPolicy; and upgrading a released 0.11.0 release with no Explorer values set changes nothing but the chart label. `tests/explorer_test.yaml` is a helm-unittest suite of twelve cases.

Version is 0.14.0 because it stacks behind #832 and #833. If those do not land first it should come down.

Closes #831.


charts/influxdb3-enterprise/templates/networkpolicy.yaml
```diff
@@ -481,3 +492,64 @@ spec:
...
+            matchLabels:
+              {{ .Values.networkPolicy.ingress.ingressController.namespaceLabel }}: {{ .Values.networkPolicy.ingress.ingressController.namespace }}
+          podSelector:
+            matchLabels:
+              {{ .Values.networkPolicy.ingress.ingressController.podLabelKey }}: {{ .Values.networkPolicy.ingress.ingressController.podLabelValue }}
+      ports:
+        # NetworkPolicy matches the destination pod port, not the Service port
+        - protocol: TCP
+          port: 8080
+    {{- end }}
```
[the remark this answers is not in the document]
tomklapka: The premise does not hold, but the rendering was untidy so I changed it anyway.

`ingress: null` is valid against the NetworkPolicy schema - the field is nullable - and a server-side dry run against a 1.31 API server accepts the object: `networkpolicy.networking.k8s.io/...-explorer created (server dry run)`. kubeconform passes it too. What it actually means is deny-all inbound, since Ingress is in policyTypes and the rule list is empty, which is arguably what turning `fromIngressController` off asks for. The querier policy avoids the case only because it still has its inter-component rule.

It now renders an explicit `ingress: []` with a comment saying what that does, rather than a bare key with a dangling comment under it. Same semantics, no longer reads like a bug.

tomklapka:
Dropped the integrated Web UI from this PR. The standalone Explorer stays and is the only path the chart offers now. Reasoning, since this removes something the original PR added.

The chart only ever produces split roles. The four workloads render `--mode=ingest`, `--mode=query`, `--mode=compact` and `--mode=process`; there is no `all` mode and no value that changes one, so as the chart stands the embedded UI can only land on a query-only node.

Writing does not work there. Explorer's documented feature set includes writing - the get-started guide is connect, write, query - but a write against a query-only node returns `cannot write to a read-only server: must enable ingest mode`, verified through the UI's own API. Pointing it at an ingester only swaps the problem: a query there returns `Node mode without query cannot retrieve window execs`.

That half is fixable, and I want to correct my own earlier framing of it. The modes combine. A node started as `--mode=query,ingest,webui` comes up as `mode=[Query, Ingest, Webui]`, and on kind a line-protocol write to it returns 204 and reads back. It needs no extra storage either - the ingester and querier StatefulSets already mount the same four volumes. So this is a chart change, not a product limit. It is simply not expressible today: each workload's mode is a literal in its template and the chart has no extraArgs anywhere.

The other two limits do not move with the roles.

It cannot be preconfigured, which is #830, and its server configuration is scoped to the browser session. In the guest that ships with 3.11.2 every read and write of a server configuration filters on the session id (`api/configurations/handlers.rs`) and there is no user or login concept. The cookie lives 120 days, so one person in one browser configures once - but a colleague, another browser or a cleared cookie starts again at the first-run screen and pastes an admin token. The standalone re-seeds itself on every new session; I checked the two side by side.

Enabling it also puts an unauthenticated surface on the querier's port: the UI takes every path the InfluxDB API does not claim, and that dispatch runs ahead of the auth layer, so paths that returned 401 begin answering without a token.

So the question is not whether the embedded UI can be made to work. It is whether a chart change plus those two caveats is worth carrying a second UI that the standalone already covers. I landed on no, but that is a judgement rather than a fact, and it removes a feature from someone else's PR - say so if you read it the other way.

The standalone Explorer does the whole job when it has a combined endpoint to point at, and that is a real condition rather than a detail. Explorer holds one active connection at a time - it can store several servers and switch between them, but a write and a query in the same sitting both go to whichever is active, and preconfiguration seeds exactly one. So the endpoint has to route both.

The chart's ingress is such an endpoint: `/api/v2/write` and `/api/v3/write_lp` reach the ingesters, the query paths reach the queriers, and a write from the Explorer through it lands and reads back on a cluster. `explorer.defaultConnection.server` now defaults to `ingress.host` when the ingress is enabled, rather than to the querier Service as before, because that default was read-only and said nothing about it.

Anyone routing another way - Gateway API, a service mesh, an OpenShift Route, a load balancer, an Ingress managed outside this chart - sets that value to their own address. Without any combined endpoint the Explorer is read-only, which the README now states plainly instead of leaving it to be discovered.

If the embedded UI later gains preconfiguration, the wiring is in this PR's history and easy to restore.

charts/influxdb3-enterprise/templates/networkpolicy.yaml
```diff
@@ -481,3 +492,68 @@ spec:
...
+    # The Explorer reaches InfluxDB over HTTP like any other client
+    - to:
+        - podSelector:
+            matchLabels:
+              {{- include "influxdb3-enterprise.selectorLabels" . | nindent 14 }}
+              app.kubernetes.io/component: querier
+      ports:
+        - protocol: TCP
+          port: 8181
```
bednar, line 557: [P2] Allow Explorer egress to the configured connection endpoint

With networkPolicy.enabled=true and the default policyTypes including Egress, the Explorer can only reach DNS and this release's querier pods on TCP 8181. However, explorerDefaultConnectionServer selects ingress.host whenever ingress.enabled=true, so the Explorer must reach the ingress controller instead. This already occurs with ci/explorer-values.yaml: the generated connection points to http://influxdb.ci.example.com:8181, but the policy has no egress rule allowing that destination. On a cluster enforcing NetworkPolicy, the configured connection times out; explicitly configured external endpoints are likewise blocked.

Please provide configurable egress rules for the actual connection endpoint, cover the chart-managed ingress route, and add a regression test for ingress plus NetworkPolicy being enabled together.

bednar (changes requested):
One functional issue needs addressing: the Explorer egress policy blocks the ingress endpoint selected by its default connection. See the inline comment. Helm lint and rendering with ci/explorer-values.yaml passed; the mismatch was confirmed in the rendered manifests. No runtime test on a NetworkPolicy-enforcing cluster was performed.

charts/influxdb3-enterprise/templates/networkpolicy.yaml
tomklapka: Fixed. With the chart's ingress on, the Explorer may reach the ingress controller on its HTTP and HTTPS pod ports (`networkPolicy.ingress.ingressController.ports`, 80/443 by default, so the ingress-nginx webhook stays closed), and `explorer.networkPolicy.extraEgress` covers other endpoints. `ingress.port` has to be the port the ingress is reachable on, which the README now says. Checked on kind with NetworkPolicy enforced: the default connection timed out before and gets 200 through the ingress after. ddf43d4

bednar (dismissed):
The previously reported Explorer egress issue is addressed. The policy now allows access to the ingress controller, with configurable controller ports and extra egress rules for other endpoints. Reviewed the updated templates and rendered policy; all 15 Explorer unit tests and Helm lint passed, and both CI checks are green. No further change requests. A runtime cluster test was not repeated during this review.

tomklapka:
Merged master and renumbered the chart from 0.14.0 to 0.12.0 so this can land first. Nothing else changed since your approval, but the push dismissed it. 117fc75

charts/influxdb3-enterprise/templates/ingress-explorer.yaml
```diff
@@ -0,0 +1,41 @@
...
+    {{- range .Values.explorer.ingress.tls }}
+    - hosts:
+        {{- range .hosts }}
+        - {{ . | quote }}
+        {{- end }}
+      secretName: {{ .secretName }}
+    {{- end }}
+  {{- end }}
+  rules:
+    - host: {{ .Values.explorer.ingress.host | quote }}
```
bednar, line 31: Anyone who can reach this Ingress can use the configured Explorer connection with the configured API token’s permissions. With the documented `mode: admin` and an admin token, that can grant full database administration; Explorer does not add ingress authentication. Please document this exposure prominently and direct operators to restrict or authenticate the ingress and use a least-privilege token before exposing it.

bednar (changes requested):
Please address the inline security finding on the Explorer Ingress: document that anyone who can reach it can use the configured connection with the API token permissions, and require ingress restrictions/authentication plus a least-privilege token.

charts/influxdb3-enterprise/templates/ingress-explorer.yaml
tomklapka: Documented in three places: a warning in the Explorer section of the README with an ingress-nginx auth example and least-privilege token guidance, comments on `explorer.ingress` and `defaultConnection.apiToken` in values.yaml, and an install note shown only when a preconfigured Explorer is exposed through its ingress or a non-ClusterIP Service. Also merged master, so the chart is 0.14.0 now. 2667ac2

bednar (approved):
Reviewed the latest commit 2667ac2; the reported Explorer exposure risk is documented and the checks are green.

tomklapka:
Closing this. The decision on the product side is webui only: the Web UI built into the server binary, not the standalone Explorer image. Shipping both would also leave a version mismatch, standalone on 1.9.0 against the 1.10.1 guest the server carries.

The webui needs almost none of what is here, so starting over is cleaner than reworking it. It is a mode on the querier plus a session secret, not a Deployment with its own Service, PVC, secrets and ingress. The NetworkPolicy rules and the token notes in the README will carry over.

A draft for the webui comes next, based on the sources on influxdb_pro main, so the chart is ready when the release with intra-cluster routing and login lands.

